PR #8, opened 2021-11-03
knight evaluation

mosheraboh:


alex-golts (approved):
I reviewed the Brier Score and Brier Skill Score implementation

PR #16, opened 2021-11-21
some codes and simplify

abnsy:


mosheraboh (approved):
Thanks!
Reviewed the code.
All the changes I've reviewed are code style modifications and there is no change in behavior.
Just wanted to double-check it, if that is the case go ahead and merge it.

PR #20, opened 2021-12-17
KNIGHT task 2 baseline

alex-golts:


fuse_examples/classification/knight/baseline/dataset.py
```diff
@@ -71,7 +71,14 @@ def prepare_clinical(sample_dict: dict) -> dict:
 
 def knight_dataset(data_dir: str = 'data', cache_dir: str = 'cache', split: dict = None, \
     reset_cache: bool = False, post_cache_processing_func: Optional[Callable] = None, \
-        rand_gen = None, batch_size=8, resize_to=(256,256,110)):
+        rand_gen = None, batch_size=8, resize_to=(256,256,110), task_num=1):
+    if task_num == 1:
+        target_name='data.gt.gt_global.task_1_label'
```
mosheraboh, line 76: Both task_num and target_name are defined in two different places - can we define all the per task variables in one place and pass it

mosheraboh (approved):
Great!
Just one comment inline

fuse_examples/classification/knight/baseline/dataset.py
alex-golts: I think I addressed it now

mosheraboh: I meant to write num_classes and target_name.
After your last modifications. You still define num_classes and target_name twice, right?

alex-golts: I see what you mean. please check now

mosheraboh (approved):
Great! Thanks!

PR #22, opened 2021-12-26
Clinical processor improvements

alex-golts:


fuse_examples/classification/knight/baseline/clinical_processor.py
```diff
@@ -44,16 +43,61 @@ def __init__(self, json_filename: str, columns_to_tensor: Optional[Union[List[st
                         When None (default) no columns are converted.
         """
         self.json_filename = json_filename
-        self.data = create_knight_clinical(json_filename)
+        self.data = pd.read_json(json_filename)
```
mosheraboh, line 46: I missed it before.
Since we might process the sample with multiple processes, it's better to hold the data in dictionary rather than dataframe.
Can you convert it to a dictionary here? (look at FuseProcessorDataframe for reference) - it should improve the running time.

fuse_examples/classification/knight/baseline/clinical_processor.py
```diff
@@ -44,16 +43,61 @@ def __init__(self, json_filename: str, columns_to_tensor: Optional[Union[List[st
...
+            items.smoking_history.iloc[0] = 1
+        elif items.smoking_history.iloc[0] == 'current_smoker':
+            items.smoking_history.iloc[0] = 2
+
+        if items.last_preop_egfr.iloc[0]['value'] == '>=90':
+            items.last_preop_egfr.iloc[0] = 90
+        else:
+            items.last_preop_egfr.iloc[0] = items.last_preop_egfr.iloc[0]['value']
+
+        # Task 1 labels:
```
mosheraboh, line 82: This should go the GT processor only, right?
I guess that already have it in your GT processor

mosheraboh (commented):
Thanks Alex!
Two comments inline.

fuse_examples/classification/knight/baseline/clinical_processor.py
alex-golts: done. please check

fuse_examples/classification/knight/baseline/clinical_processor.py
alex-golts: I created now a separate GT processor, so now each processor only uses its corresponding data. please review. earlier both GT and input data used the same processor

mosheraboh (approved):
Great, Thanks!

PR #24, opened 2022-01-10
prepare for test data, change scripting convention

alex-golts:


mosheraboh (approved):
Great! Thanks Alex!
Can please also add to knight README explanation about test set 
(mention that a link will be sent to knight participants only at this point), also how to move from validation inference to test inference.

PR #26, opened 2022-01-18
support for test data and optional evaluation

alex-golts:


fuse_examples/classification/knight/baseline/clinical_processor.py
```diff
@@ -76,11 +76,14 @@ def __call__(self, sample_desc: Hashable):
         elif sample_data['smoking_history'] == 'current_smoker':
             sample_data['smoking_history'] = 2
 
-        if sample_data['last_preop_egfr']['value'] == '>=90':
-            sample_data['last_preop_egfr'] = 90
+        if sample_data['last_preop_egfr'] is None:
+            sample_data['last_preop_egfr'] = -1 
         else:
-            sample_data['last_preop_egfr'] = sample_data['last_preop_egfr']['value']
```
mosheraboh, line 82: Do you know what is the distribution of this value?
Not sure if -1 is a good idea here

fuse_examples/classification/knight/make_predictions_file.py
```diff
@@ -64,7 +64,7 @@ def make_predictions_file(model_dir: str,
             # For this example, we use split 0 out of the the available cross validation splits
             split = split[0]
     if split is None: # test mode
-        json_filepath = os.path.join(data_path, 'knight_test.json')
+        json_filepath = os.path.join(data_path, 'features.json')
```
mosheraboh, line 67: features.json is the file with the labels?

fuse_examples/classification/knight/baseline/clinical_processor.py
alex-golts: -1 is out of the normal range. do you propose to use the median or something like that instead?

fuse_examples/classification/knight/make_predictions_file.py
alex-golts: no, this is the input features file. the name got changed from last time

mosheraboh (approved):

mosheraboh (approved):
Great, thanks Alex.

PR #29, opened 2022-01-30
knight - make targets for test set

mosheraboh:


fuse_examples/classification/knight/baseline/dataset.py
```diff
@@ -121,19 +121,26 @@ def knight_dataset(data_dir: str = 'data', cache_dir: str = 'cache', split: dict
...
         else:
             gt_processors = {}
-    # we use the same processor for the clinical data and ground truth, since both are in the .csv file
-    # need to make sure to discard the label column from the data when using it as input
-    input_processors = {
-        'image': KiTSBasicInputProcessor(input_data=image_dir, resize_to=resize_to),
-        'clinical': KiCClinicalProcessor(json_filename=json_filepath)
-    }
+
+    if only_labels:
```
alex-golts, line 132: how will this support the "test" mode where we do have the test labels but also want to predict (i.e use the images)?
maybe we should have two parameters? test_labels AND labels_only?

fuse_examples/classification/knight/make_targets_file.py
```diff
@@ -56,25 +56,25 @@ def make_targets_file(data_path: str,
...
+    else:
+        raise Exception(f"Error: unexpected split format {split}")
+    
+    _, validation_dl, test_dl, _, _, _  = knight_dataset(data_path, cache_path, split, reset_cache=False, only_labels=True)
 
     # Export targets
-    if test_labels:
-        targets_df = DatasetExport.export_to_dataframe(test_dl.dataset, ["data.descriptor", "data.gt.gt_global.task_1_label", "data.gt.gt_global.task_2_label"])
-    else:
+    if isinstance(split, str):
```
alex-golts, line 73: maybe better to leave the condition "if only_labels" as it's more explicit and easy to understand

alex-golts (commented):
Thanks!
Please see two comments inline...

fuse_examples/classification/knight/make_targets_file.py
mosheraboh: In both cases, it's only labels.
If --> validation labels
else --> test lables

fuse_examples/classification/knight/baseline/dataset.py
mosheraboh: We don't support it - and we didn't support it in the previous version of the code as well.
images only mode is for inference and labels only mode is for evaluation.
Does it seem important to you to support?

PR #30, opened 2022-02-06
changed model_default to accept either conv_inputs (for convolutional…

floccinauc:
changed model_default to accept either conv_inputs (for convolutional backbone) of backbone_args (for generic backbone).
The new model_default __init__ throws an exception if neither conv_inputs nor backbone_args are set, and if both are set.


mosheraboh (approved):
Great! Thanks!

PR #37, opened 2022-02-20
mri examples

taltlusty:
1. fix maxpool applied to (1,x,y) layers -> stride (1,2,2) 
2. num of workers ->0
3. add requirements for mri examples


mosheraboh (changes requested):
Thanks Tal,
Just one think about the num_workers inline.

fuse_examples/classification/duke_breast_cancer/dataset.py
```diff
@@ -161,7 +161,7 @@ def duke_breast_cancer_dataset(paths,train_common_params,lgr):
                                          )
 
     lgr.info(f'- Load and cache data:')
-    train_dataset.create()
+    train_dataset.create(num_workers=0)
```
mosheraboh, line 164: You shouldn't set the num_workers here (in train set) to 0. It should work as-is. Can you please change it back?

fuse_examples/classification/prostate_x/dataset.py
```diff
@@ -174,7 +174,7 @@ def prostate_x_dataset(paths,train_common_params,lgr):
     gpu_ids_for_caching = []
     lgr.info(f'- Load and cache data:')
 
-    train_dataset.create()
+    train_dataset.create(num_workers=0)
```
mosheraboh, line 177: Same here, can you please change it back?

mosheraboh (changes requested):
Thanks Tal! looks good.
We just need to change back the num workers in train set, see inline.

mosheraboh (approved):

PR #52, opened 2022-04-05
move generated env to dedicated path

mosheraboh:


run_all_unit_tests.sh
```diff
@@ -15,6 +15,7 @@ lockfailed()
 # create an environment including the set of requirements specified in fuse-med-ml/requirements.txt (if not already exist)
 create_env() {
     force_cuda_version=$1
+    dest_path=$2
```
YoelShoshan, line 18: you don't seem to use this var, is that intended?

YoelShoshan, line 18: maybe I misunderstand, but isn't there a confusion here between "dest_path" and "env_path" ?

YoelShoshan (approved):
LGTM

PR #53, opened 2022-04-07
Merge fuse2 utils

mosheraboh:


fuse/utils/config_tools/__init__.py
```diff
@@ -0,0 +1,4 @@
+from fuse.utils.config_tools.config_tools import (
```
YoelShoshan, line 1: something to think about - generally about configs:

we might want to switch into using:
1. Hydra configs - AFAIK they require decorating your "main" function (and I don't know if they can be used to additionally decorate other functions)
the advantages is that this lib:
* Allow building configs hierarchically 
* Allows overriding params with CLI
* Possibly plays better with some HP search tools, which build on being able to use CLI

2. yaml based configs - maybe that's ok and can be considered as just another implementation of the same API that we defined

YoelShoshan (approved):
Looks good!
a minor comment inline

PR #54, opened 2022-04-07
A unit test for the hello world notebook.

SagiPolaczek:
* Using the module 'testbook'
* Currently only testing the result's accuracy 
* TODO: Added more tests


SagiPolaczek:
I've used an external module that I found: 'testbook' for testing jupyter notebooks. On the LSF with my conda env it works good.
The test is failing on the remote because the kernel 'python3' is missing.
While looking for a solution I've encounter [this one](https://stackoverflow.com/questions/38257138/jupyter-no-such-kernel-named-python3). Is this relevant?

Another option for unit testing the notebook is to run it as a script. But I think the method in the PR is much cleaner and will be useful for further notebooks.

I would like to hear your thoughts please @mosheraboh

mosheraboh (commented):
Any chance that adding ipykernel to the list of requirements will solve it?

fuse/managers/manager_default.py
```diff
@@ -1055,6 +1055,17 @@ def _handle_dataset_summaries(self, train_dataloader: DataLoader, validation_dat
             self.logger.info(dataset_summary)
         pass
 
+    def set_device(self, device: str):
+        """
+        set the manger's device to a given one.
+        :param device: device to set
+        """
+        train_params = {'device' : device}
```
mosheraboh, line 1063: update the state directly. (self.state.device).

fuse_examples/tests/test_notebook_hello_world.py
```diff
@@ -0,0 +1,16 @@
+import os
+import unittest
+from testbook import testbook
+
+class NotebookHelloWorldTestCase(unittest.TestCase):
+    
+    def test_example_2(self):
+        # Execute the whole notebook and save it as an object
+        with testbook('fuse_examples/tutorials/hello_world/hello_world.ipynb', execute=True, timeout=600) as tb:
```
mosheraboh, line 9: Isn't the first cell install fuse-med-ml? Should we skip it?

fuse_examples/tests/test_notebook_hello_world.py
```diff
@@ -0,0 +1,16 @@
+import os
+import unittest
+from testbook import testbook
+
+class NotebookHelloWorldTestCase(unittest.TestCase):
+    
+    def test_example_2(self):
```
mosheraboh, line 7: The test takes a long time.
It might be the CPU - maybe the default should be the GPU0 (see multimodal notebook) and keep in comment the steps to move it to CPU?

requirements.txt
```diff
@@ -24,4 +24,7 @@ nibabel
 pycocotools>=2.0.1
 xmlrunner
 paramiko
+testbook
+termcolor
```
mosheraboh, line 28: we already have it in the list

mosheraboh (changes requested):
Great,
testbook will be very helpful looking forward.
Please address the comment inline and merge to master when ready.

fuse_examples/tests/test_notebook_hello_world.py
SagiPolaczek: It's not something that we want to test? Because it's already been testing in the Jenkins pipeline itself?

I think it won't be an issue to skip it tho.

fuse_examples/tests/test_notebook_hello_world.py
SagiPolaczek: Indeed long time. I'll change it

fuse_examples/tests/test_classification_mnist.py
```diff
@@ -58,6 +58,7 @@ def test_template(self):
         self.assertGreaterEqual(results['metrics.auc.macro_avg'], threshold)
 
     def tearDown(self):
+        FuseUtilsGPU.deallocate_gpu()
```
SagiPolaczek, line 61: @mosheraboh 
Do we want to delete that line? 
It was added in the debug session we had, but I'm not sure it's necessary.

run_all_unit_tests.py
```diff
@@ -32,7 +32,7 @@ def mehikon(a,b):
     suite = None
     for curr_subsection in sub_sections:
         curr_subsuite = unittest.TestLoader().discover(
-            f'{search_base}/{curr_subsection}', 'test*.py', top_level_dir=search_base
+            f'{search_base}/{curr_subsection}', 'test_*.py', top_level_dir=search_base
```
SagiPolaczek, line 35: @mosheraboh 
Same as before :)

mosheraboh: Yes, revert it.

fuse_examples/tests/test_classification_mnist.py
mosheraboh: We can keep it until we resolve the issue

mosheraboh (dismissed):

fuse_examples/tutorials/hello_world/hello_world.ipynb
```diff
@@ -47,9 +47,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "!git clone https://github.com/IBM/fuse-med-ml.git\n",
-    "%cd fuse-med-ml\n",
-    "!pip install -e ."
+    "install_fuse = False # change to 'True' to clone and install fuse-med-ml. \n",
```
SagiPolaczek, line 50: @mosheraboh 
Yesterday I've opened a pull request on testbook's repo adding a 'skip' feature 🙂 
When they'll approve it I'll be able to clean it.

PR #62, opened 2022-04-18
Mnist fuse2 style

mosheraboh:


examples/fuse_examples/imaging/classification/mnist/runner.py
```diff
@@ -129,38 +134,26 @@ def run_train(paths: dict, train_params: dict):
...
+    sampler = BatchSamplerDefault(dataset=train_dataset,
                                        balanced_class_name='data.label',
                                        num_balanced_classes=10,
                                        batch_size=train_params['data.batch_size'],
                                        balanced_class_weights=None)
     lgr.info(f'- Create sampler: Done')
 
     # Create dataloader
-    train_dataloader = DataLoader(dataset=train_dataset, batch_sampler=sampler, num_workers=train_params['data.train_num_workers'])
+    train_dataloader = DataLoader(dataset=train_dataset, batch_sampler=sampler, collate_fn=CollateDefault(), num_workers=train_params['data.train_num_workers'])
```
alex-golts, line 147: why is it now not Pytorch's default collate function? won't it work as well?

alex-golts (commented):
Just one question about the need for explicit definition of fuse default collate function, that wasn't required before in this use case

PR #63, opened 2022-04-18
Unittest for hello world notebook

SagiPolaczek:
The PR isn't ready yet. I've just applied the same changes.

### DIFF
--- a/fuse/dl/managers/manager_default.py
+++ b/fuse/dl/managers/manager_default.py
@@ -945,6 +945,16 @@ class ManagerDefault:
             self.logger.info(dataset_summary)
         pass
 
+    def set_device(self, device: str):
+        """
+        set the manger's device to a given one.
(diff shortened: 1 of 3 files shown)


fuse/dl/managers/manager_default.py
```diff
@@ -945,6 +945,16 @@ def _handle_dataset_summaries(self, train_dataloader: DataLoader, validation_dat
             self.logger.info(dataset_summary)
         pass
 
+    def set_device(self, device: str):
+        """
+        set the manger's device to a given one.
+        :param device: device to set
+        """
+        train_params = {'device' : device}
+        self.set_objects(train_params=train_params)
```
mosheraboh, line 954: make sure that we don't override other arguments.

mosheraboh (changes requested):

PR #86, opened 2022-05-18
Simple op base

mosheraboh:


fuse/data/ops/op_base.py
```diff
@@ -49,14 +47,37 @@ def __call__(self, sample_dict: NDict, op_id: Optional[str], **kwargs) -> Union[
         """
         raise NotImplementedError
 
+
+    
+class OpReverseableBase(OpBase):
```
YoelShoshan, line 52: spelling -- Reversible?
Or maybe Invertible ?

fuse/data/ops/op_base.py
```diff
@@ -68,8 +89,28 @@ def reverse(self, sample_dict: NDict, key_to_reverse: str, key_to_follow: str, o
         :param key_to_follow: run the reverse according to the operation applied on this value
         :return: modified sample_dict
         """
-        raise NotImplemented
+        raise sample_dict
```
YoelShoshan, line 92: I wonder what's better as a default - exception or not.

I understand the convenience of just inherit from OpReversibleOp to make it passthrough, but I wonder if it will make users mistake and think it auto-magically actually works and inverts the op.

Perhaps it would be better to throw an exception by default, but making it more detailed - like,
"... mark it safe for reverse by blah blah if there's nothing that should be done, or implement blah blah - see doc blah blah"

fuse/data/ops/op_base.py
```diff
@@ -68,8 +89,28 @@ def reverse(self, sample_dict: NDict, key_to_reverse: str, key_to_follow: str, o
...
 
     
+def op_call(op: OpBase, sample_dict: NDict, op_id: str, **kwargs):
+    try:
+        if isinstance(op, OpReverseableBase):
+            return op(sample_dict, op_id=op_id, **kwargs)
+        else: # OpBase but note reversable
```
YoelShoshan, line 99: typo - not

YoelShoshan, line 99: should we explicitly check for OpBase in the other option, for the case that someone mistakenly provided something different? 
(and raise exception if it's neither?)

fuse/data/ops/op_base.py
```diff
@@ -68,8 +89,28 @@ def reverse(self, sample_dict: NDict, key_to_reverse: str, key_to_follow: str, o
...
+def op_reverse(op, sample_dict: NDict, key_to_reverse: str, key_to_follow: str, op_id: Optional[str]):
+    if isinstance(op, OpReverseableBase):
+        try:
+            return op.reverse(sample_dict, key_to_reverse, key_to_follow, op_id)
+        except:
+            # error messages are cryptic without this. For example, you can get "TypeError: __call__() got an unexpected keyword argument 'key_out_input'" , without any reference to the relevant op!
+            print(f'error in reverse method of op={op}, op_id={op_id}, sample_id={get_sample_id(sample_dict)} - more details above')   
+            raise 
     
+    else: # OpBase but note reversable
```
YoelShoshan, line 115: typo - not

YoelShoshan (dismissed):
Looks good!
See comments inline

fuse/data/ops/op_base.py
mosheraboh: (－‸ლ)

fuse/data/ops/op_base.py
mosheraboh: I want to keep the option to use callables  instead

PR #100, opened 2022-06-09
simplify and modify FuseDebug to fuse2

mosheraboh:


YoelShoshan (approved):
LGTM
We'll discuss manager design in general in the near future

PR #106, opened 2022-06-15
Added new ops to README files

SagiPolaczek:
ops:
OpOneHotToNumber
OpPad
OpResizeTo


SagiPolaczek:
@mosheraboh 

What is your opinion on this way of implementing 'exclude' in the configurations files?
The granularity might dynamically change when we will fix more files' formats.

mosheraboh (approved):

PR #110, opened 2022-06-17
Yoels code review

YoelShoshan:
Few code review and redesign of dl sub-package.
PR isn't ready! 
There is a test that hangs on multiprocessing (doesn't seem related so far to this PR code)


YoelShoshan:
all tests pass - ready for review

fuse/dl/models/model_wrapper.py
```diff
@@ -31,10 +31,10 @@ class ModelWrapper(torch.nn.Module):
 
     def __init__(self,
                  model: torch.nn.Module = None,
-                 model_inputs: Sequence[str] = ('data.input.input_0.tensor',),
+                 model_inputs: Sequence[str] = None, #('data.input.input_0.tensor',),
```
mosheraboh, line 34: Lets remove the default value.
Move the outputs to be the second argument (and also remove the default value)

fuse/dl/models/model_wrapper.py
```diff
@@ -31,10 +31,10 @@ class ModelWrapper(torch.nn.Module):
 
     def __init__(self,
```
mosheraboh, line 32: Maybe we should follow the dataset naming (ModelWrapSeqToDict)?

mosheraboh (dismissed):

fuse/dl/models/model_wrapper.py
YoelShoshan: done (I assume you meant 3rd argument, write here if you prefer output to be before input)

YoelShoshan: also prevent positional args, and added example values to the constructor docstring

fuse/dl/models/model_wrapper.py
YoelShoshan: renamed to ModelWrapSeqToDict - did not change the file name as we might include additional wrappers there in the future

fuse/dl/models/heads/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .common import ClassifierFCN, ClassifierFCN3D, ClassifierMLP
+from .head_1d_classifier import Head1DClassifier
```
mosheraboh, line 2: You an issue here caught in unittests with upper/smaller case in 1D/1d

mosheraboh (approved):
Seems good,
You just need to fix an issue caught in unittests, see the inline comment.

PR #111, opened 2022-06-21
Add biliear interpolation to affine augmentation + print structure ut…

egozi:
…ility fundtion to ndict


fuse/utils/ndict.py
```diff
@@ -277,6 +277,17 @@ def get_multi(self, keys: Optional[List[str]]=None) -> NDict:
         
         return ans
 
+    def print_tree(self):
+        self.print_struct(self._stored)
+
+    def print_struct(self, d, level=0):
```
mosheraboh, line 283: Can you please add type annotation + comments?
If print_struct is for internal use, please add "_" prefix.
Use a longer name than "d",

fuseimg/data/ops/aug/geometry.py
```diff
@@ -57,7 +57,8 @@ def __call__(self, sample_dict: NDict, key: str, rotate: float = 0.0, translate:
         for channel in channels:
             aug_channel_tensor = aug_input[channel].numpy()
             aug_channel_tensor = Image.fromarray(aug_channel_tensor)
-            aug_channel_tensor = TTF.affine(aug_channel_tensor, angle=rotate, scale=scale, translate=translate, shear=shear)
+            aug_channel_tensor = TTF.affine(aug_channel_tensor, angle=rotate, scale=scale, 
+                                            resample=Image.BILINEAR, translate=translate, shear=shear)
```
mosheraboh, line 61: can you please add it to the as an argument? It's ok to set bilinear as default.

mosheraboh (changes requested):
Seems good, minors inline.

mosheraboh (dismissed):
Thanks!
Please pull to resolve conflicts and then merge.

mosheraboh (approved):

PR #131, opened 2022-07-07
fuseimg extra requires

mosheraboh:


fuse/requirements.txt
```diff
@@ -6,33 +6,28 @@ pandas>=1.2
 tqdm>=4.52.0
 scipy>=1.5.4
 matplotlib>=3.3.3
-scikit-image>=0.17.2
 scikit-learn>=0.23.2
 termcolor>=1.1.0
-torch>=1.5.0,<=1.11.0
+torch>=1.5.0,<=1.11.0 # the higher limit is temporary - to avoid from error in unitests:  "Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method"
```
YoelShoshan, line 11: can you explain a bit more?
Not sure it's a good idea to limit the max pytorch version of the users...
By "temporary" do you mean it is intended to be resolved before fuse2 official release ?

fuse/requirements.txt
```diff
@@ -6,33 +6,28 @@ pandas>=1.2
...
 paramiko
 tables
 psutil
 testbook
 ipykernel
-medpy
 pytorch_lightning
 hydra-core
 omegaconf
+pycocotools>=2.0.1
```
YoelShoshan, line 29: do we want such requirements in fuse-med-ml core? or only in fuseimg ?

fuse/requirements.txt
```diff
@@ -6,33 +6,28 @@ pandas>=1.2
...
+nibabel
```
YoelShoshan, line 30: do we want such requirements in fuse-med-ml core? or only in fuseimg ?

fuse/utils/multiprocessing/run_multiprocessed.py
```diff
@@ -260,15 +277,31 @@ def inner(f: callable):
...
             except:
                 p.terminate()
                 raise
-            
+
+            if p.exception:
+               error, traceback = p.exception
+               print(f"process func {f} had an exception: {error}")
+               print(traceback)
+               raise RuntimeError(f"process func {f} had an exception: {error}")
```
YoelShoshan, line 293: great !!

YoelShoshan (approved):
please see comments inline

fuse/requirements.txt
mosheraboh: Not too much.
The last builds failed for us. All the unittest running in a child process failed with this error "Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method".
Probably something new trigger CUDA init on torch version 1.12.0. 
This is a workaround until we figure out what it is.

fuse/requirements.txt
mosheraboh: In fuseimg.
But I had to move it back here - cause currently, our core eval package also includes segmentation metrics (which use cocotools).

fuse/requirements.txt
mosheraboh: same reason as above

fuse/utils/multiprocessing/run_multiprocessed.py
mosheraboh: All rights belong to @michalozeryflato :) (I copied it with minor modifications)

fuse/requirements.txt
YoelShoshan: What about the following suggestion:
Not limit the user pytorch version, and in our unit tests use an env var that run_multiprocessed reads and actually use the 'spawn' start method ? does that make sense ?

fuse/requirements.txt
YoelShoshan: ok - I assume that this is something you want to solve after fuse2 release - we can handle it when we add fusedrug (as we'll want to minimize core dependencies then)

fuse/requirements.txt
mosheraboh: I tried it first, naively changing it to "spawn" failed (trying to serialize something).
Agree that it shouldn't be our long term solution.
It's just to release fuse2 - and not to block all active pull requests.

PR #132, opened 2022-07-07
temp remove unported examples and add CODE_OF_CONDUCT.md

mosheraboh:


SagiPolaczek (approved):
All looks ok! Thanks!

I wonder, why the DUKE's data folds have been pushed? Shouldn't we keep it locally?

PR #136, opened 2022-07-12
Fix unitests with latest pytorch

mosheraboh:
1. Fixing unitest issue: "Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method""
2. Upgrading pipeline to CUDA 11.6 to fit all GPUs in jenkins
3. Fix black and mypy version to ensure stable checkers
4. Move some dependencies to requiements_dev.txt


YoelShoshan (approved):
LGTM :)

PR #146, opened 2022-07-21
Fix bug and handle no extension files

alex-golts:


fuse/utils/file_io/path.py
```diff
@@ -12,8 +12,11 @@ def change_extension(filepath: str, new_extension: str):
     _basename = os.path.basename(filepath)
     last_dot = _basename.rfind(".")
     if last_dot < 0:
-        return filepath
-    ans = os.path.join(os.path.dirname(filepath), _basename[:last_dot] + new_extension)
+        if not add_if_no_extension:
+            return filepath
```
mosheraboh, line 16: I would expect to throw an error here or to always add if not extension (meaning no need for a flag).
(I assume you just wanted to keep backward compatibility that I am not sure required)
I am approving and leaving it for you to decide though...

mosheraboh (approved):

fuse/utils/file_io/path.py
alex-golts: yes exactly, for backward compatibility. 
I changed it to always add, as I had a use-case where that's what I needed

PR #149, opened 2022-07-25
Knight test

liamhazan:
still need to define data path in the ccc.


fuseimg/README.md
```diff
@@ -43,6 +43,8 @@ Here you can find useful ops for imaging and public datasets implementations
 * OpAugSqueeze3Dto2D - squeeze selected axis of volume image into channel dimension, in order to fit the 2D augmentation functions
 * OpAugUnsqueeze3DFrom2D - unsqueeze back to 3D, after you apply the required 2D operations
 * OpResizeTo - resize an image into given dimensions
+* OpCrop3D - crop 3d image to certain size. can be used as random crop using OpSample.
+* OpRotation3D - rotate 3d image across the 3 planes xyz.
```
mosheraboh, line 47: Add below your dataset.

mosheraboh (commented):
Thanks Liam!
Please merge master into your branch - currently, I see old changes as well.
See my two comments inline.
I also didn't find the test for the training process. See 
fuse-med-ml/examples/fuse_examples/tests/test_classification_stoic21.py as an example.

fuseimg/datasets/knight.py
```diff
@@ -245,12 +242,7 @@ def dataset(
         cache_dir: str = "cache",
         split: dict = None,
```
mosheraboh, line 243: How should a user generate a split?
I think that it's better to get sample_ids as the other examples and to include all samples if it was set to None.

mosheraboh (changes requested):

fuseimg/datasets/knight.py
liamhazan: there is a splits pkl file which contains 5 different splits which are necessary to compare results.

mosheraboh: It's ok to use it for the example you are implementing.
But the dataset implementation needs to be more general, so we shouldn't ask the users to create their own split file to use it.
And if we do, you need to carefully explain the structure and how to create a new one.

mosheraboh (changes requested):
The test will probably fail - you should use environment variable instead of local path

mosheraboh (approved):

PR #151, opened 2022-07-27
implement basic ops debug

mosheraboh:


fuse/data/ops/ops_debug.py
```diff
@@ -0,0 +1,110 @@
...
+    ```
+    (OpPrintKeys(first_sample_only), dict()),
+    ```
+    """
+
+    def __call__(self, sample_dict: NDict) -> Any:
+        if self.should_debug_sample(sample_dict):
+            print(f"Sample {get_sample_id(sample_dict)} keys:")
+            for key in sample_dict.keypaths():
+                print(f"{key}")
```
SagiPolaczek, line 55: Suggestions:
1. Use `print(f"{key}", end = " ")`can be better in case there are a lot of keys. Not sure how it is common, but if there are 20+ keys printing each key in new line could cause a mess. I think thats the case int he MRI right now.. what do you think?
2. Also we can print the total number of keys.

fuse/data/ops/ops_debug.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        :param sample_ids: apply for the specified sample ids. To apply for all set to None.
+        :param first_sample_only: apply for the first sample only
+        """
+        super().__init__()
+        self._sample_ids = sample_ids
+        self._first_sample_only = first_sample_only
+        self._first_sample_done = False
+
+    def should_debug_sample(self, sample_dict: NDict) -> bool:
```
SagiPolaczek, line 22: Suggestion:
Added an option to limit the amount of samples to a number (and not only by names) in case the user want to debug 5 random samples (without mentioning the names).
I can see it useful for the `OpVisImage`.  (In case the user want to examine 3-5 random images in a single run).

SagiPolaczek (approved):
Thanks Moshiko! Looks great!

I added some comments inline, most of are suggestions that came up to my mind while reading the Ops.
For sure I'll use some of those Ops when I'll come back after the exam 😎 😆  

Also I think it should be a unittest that run a single pipeline that uses all the ops. It might be a solution for a quick test just to check that nothing brakes. I can implement that next week of course.

fuse/data/ops/ops_debug.py
```diff
@@ -0,0 +1,110 @@
...
+    It's recommended, but not a must, to run it in a single process.
+    Add at the top your script to force single process:
+    ```
+    from fuse.utils.utils_debug import FuseDebug
+    FuseDebug("debug")
+    ```
+
+    Example:
+    ```
+    (OpPrintKeys(first_sample_only), dict()),
```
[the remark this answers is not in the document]
mosheraboh: Right, thanks!

fuse/data/ops/ops_debug.py
```diff
@@ -0,0 +1,110 @@
...
+    ```
+    """
+
+    def __call__(self, sample_dict: NDict) -> Any:
+        if self.should_debug_sample(sample_dict):
```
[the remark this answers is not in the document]
mosheraboh: It's possible - but then the user will need to implement another function and not the __call__ function.
Originally, I thought that I prefer just to provide the should_debug_sampe() function.
But, I think I've changed my mind.
I will push tomorrow another option.

fuse/data/ops/ops_debug.py
mosheraboh: I think that newline for every key is easier to follow (they are also kind of sorted - so it is easier to find the key).

fuse/data/ops/ops_debug.py
mosheraboh: Good idea! I will change `first_sample_only` to `num_samples_to_debug`

fuse/data/ops/ops_debug.py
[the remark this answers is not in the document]
SagiPolaczek: COOL Thanks :)

fuse/data/ops/ops_debug.py
```diff
@@ -0,0 +1,125 @@
...
+from fuse.utils import NDict
+from fuse.data import OpBase
+import numpy
+import torch
+
+
+class OpDebugBase(OpBase):
+    """
+    Base class for debug operations.
+    Provides the ability to limit the samples to the debug.
```
YoelShoshan, line 13: -> to limit samples to debug

YoelShoshan: maybe mention that first K samples will be debugged and it will be passthrough for the rest?

YoelShoshan (dismissed):
LGTM

PR #152, opened 2022-07-30
* added remove_extension, changed default deepdiff to ignore nans, addeds proper implementation of keys() values() and items() to NDict, tried to also highlight more the faulting op in pipeline ops error

YoelShoshan:
* Added "remove_extension" to path utils
* Changed default deepdiff behavior to ignore nans in comparison, 
* Added keys() items() and values() to our NDict  (until now it returned empty iterables for those which is incorrect)
* Tried to highlight more the faulting op in pipeline ops error


fuse/utils/file_io/path.py
```diff
@@ -28,3 +28,26 @@ def get_extension(filepath: str) -> str:
...
+    For example - remove_extension('/a/b/c/d/asdf.txt') with return '/a/b/c/d/asdf'
+    """
+
+    _dirname = os.path.dirname(filepath)
+    _basename = os.path.basename(filepath)
+    if '.' not in _basename:
+        return filepath
+
+    last_dot = _basename.rfind(".")
+    if last_dot < 0:
```
mosheraboh, line 45: It can't happen

fuse/utils/file_io/path.py
```diff
@@ -28,3 +28,26 @@ def get_extension(filepath: str) -> str:
...
+        return ""
+    ans = os.path.join(_dirname, _basename[:last_dot])
+    return ans
+
+
+if __name__ == '__main__':
```
mosheraboh, line 51: remove

mosheraboh (approved):

fuse/utils/file_io/path.py
YoelShoshan: why not? it represents the case that there is no '.' at all in the filename (and therefore no extension to remove)
for example - /a/b/c/foo

-1 just marks that it didn't find it

In [1]: 'abcd'.rfind('z')
Out[1]: -1

fuse/utils/file_io/path.py
YoelShoshan: done

fuse/utils/file_io/path.py
mosheraboh: Just because you already checked it in line 41 and returned the original filepath in such a case.

YoelShoshan: oh I see! so it's redundant, will fix - thanks!

YoelShoshan: done

mosheraboh (dismissed):

mosheraboh:
LGTM! Thanks!

mosheraboh (approved):

PR #155, opened 2022-08-08
Ukbb mri neck to knee ozery

michalozeryflato:


fuse/data/datasets/dataset_default.py
```diff
@@ -40,6 +40,7 @@ def __init__(
         dynamic_pipeline: Optional[PipelineDefault] = None,
         cacher: Optional[SamplesCacher] = None,
         allow_uncached_sample_morphing: bool = False,
+        always_return_dict = True,
```
mosheraboh, line 43: let's changed the allow_uncached_sample_morphing to mode instead of boolean.
We can have a mode that goes over all the samples to filter Nones or to split samples.
And a mode that skips it and returns whatever the pipeline generates.
See if you can come up with good names for the argument and the modes.

fuse/utils/data/collate.py
```diff
@@ -100,7 +100,7 @@ def _collect_values_to_list(self, samples: List[str], key: str) -> Tuple[List, b
                 if self._raise_error_key_missing:
                     raise Exception(f"Error: key {key} does not exist in sample {index}: {sample}")
                 else:
-                    value = None
```
mosheraboh, line 103: I thought that your use-case deals with a sample that is None and missing a key in one of the samples.
Cann you comment about it?

mosheraboh (changes requested):
Thanks, Itai!
I should review just the files if fuse core (under fuse dir), right?
I think that those modifications can be very useful - let's make them a bit more generic though.

PR #169, opened 2022-08-25
remove opencv version limitation

mosheraboh:


fuseimg/requirements.txt
```diff
@@ -3,6 +3,6 @@
 scikit-image>=0.17.2
 torchvision>=0.8.1
 SimpleITK>=1.2.0
-opencv-python<=4.3.0.36 
+opencv-python>4.3.0.38
```
vagenas, line 6: @mosheraboh, can you explain why `>4.3.0.38` is used, i.e. an exclusive condition and that particular version?
I would rather expect `>=4.2.0.32`, directly associated to [CVE-2019-5064](https://github.com/advisories/GHSA-q799-q27x-vp7w).

mosheraboh: We had an issue with 4.3.0.37 & 4.3.0.38 (in fuse1) and I can't reproduce them anymore.
I will follow your suggestion then and revisit if we will have a similar issue in the future.

vagenas (approved):
Approved for addressing the [CVE-2019-5064](https://github.com/advisories/GHSA-q799-q27x-vp7w) vulnerability issue.

@mosheraboh I would actually like us to review the dependency management mechanism (e.g. the interaction between the setuptools configuration –currently `setup.py`– and the various `requirements.txt` files), but let us do this separately without blocking this fix.

mosheraboh:
Thanks @vagenas,  would be happy to talk about the dependency management mechanism.

vagenas:
FYI @mosheraboh, created an issue for the dependency management point: https://github.com/BiomedSciAI/fuse-med-ml/issues/212

Issue #172, opened 2022-09-02
Example of fuse with knight dataset does not work with cpu


agiova:
**Describe the bug**\
Example folder: `examples/fuse_examples/imaging/classification/knight/baseline`
After setting the `config.yaml` with `num_gpus : 0`, this error follows:
```
Failed to run nvidia-smi
Traceback (most recent call last):
  File "/[..]/fuse-med-ml/examples/fuse_examples/imaging/classification/knight/baseline/fuse_baseline.py", line 221, in <module>
    main(config_path)
  File "/[..]/fuse-med-ml/examples/fuse_examples/imaging/classification/knight/baseline/fuse_baseline.py", line 106, in main
    GPU.choose_and_enable_multiple_gpus(cfg["num_gpus"], force_gpus=None)
  File "/[..]/fuse-med-ml/fuse/utils/gpu.py", line 74, in choose_and_enable_multiple_gpus
    raise Exception("could not auto-detect available GPUs")
Exception: could not auto-detect available GPUs
```

**FuseMedML version**\
commit https://github.com/BiomedSciAI/fuse-med-ml/commit/6a90bf3af9ca3724ae2882b717bc134ac3a930e3

**Python version**\
Python 3.9.13

**To reproduce**\
Following steps:
- Modify `config.yaml` with `num_gpus : 0`
- (set env variables)
- Execute `examples/fuse_examples/imaging/classification/knight/baseline/fuse_baseline.py`

**Expected behavior**\
Script should not search GPUs if `config.yaml` with `num_gpus : 0`.

agiova:
fixed with PR #173

PR #173, opened 2022-09-03
fix (#172): support cpu in knight example

agiova:
Adding conditional configuration of lightning trainer depending on cpu or gpu, fixes issue #172


mosheraboh (approved):
Thanks!

PR #177, opened 2022-09-13
Lightweight mode in DatasetDefault which alllows good GPU utilization in large datasets (e.g. 100M samples)

YoelShoshan:


fuse/utils/ndict.py
```diff
@@ -308,6 +308,14 @@ def _print_tree_static(data_dict: dict, level: int = 0) -> None:
                 print("---" * level, key)
                 NDict._print_tree_static(data_dict[key], level)
 
+    def describe(self) -> None:
+        for k in self.flatten().keys():
```
mosheraboh, line 312: self.keypaths()

fuse/data/ops/caching_tools.py
```diff
@@ -142,3 +146,26 @@ def __init__(self, blah, blah2):
         del curr_stack
 
     return str_desc
+
+
+# TODO: consider adding "ignore list" of args that should not participate in cache value calculation (for example - "verbose")
+def run_cached_func(cache_dir: str, func: Callable, *args, **kwargs):
```
mosheraboh, line 152: What is it? are you using it?

mosheraboh (dismissed):
Looks good.
Two minors inline.

fuse/data/ops/caching_tools.py
YoelShoshan: adding docstring

YoelShoshan: copying it here for reference:

`def run_cached_func(cache_dir: str, func: Callable, *args, **kwargs) -> Any:
    """
    Will cache the function output in the first time that
     it is executed, and will load from cache on the next times.

    The cache hash value will be based on the function name, the args, and the code of the function.

    Args:

    :param cache_dir: the directory into which caches will be stored/loaded
    :param func: the function to run
    :param *args: positional args to provide to the function
    :param **kwargs: kwargs to provide to the function

    """`

mosheraboh (approved):

PR #179, opened 2022-09-19
Support Python 3.10

SagiPolaczek:
Adding support to `python 3.10`.
NOTE that the CI python environment is set to 3.10 .

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ pipeline {
             steps {
                 sh '''
                 rm -rf test-reports/TEST-*.xml
-                jbsub -wait -out ccc_log.txt -queue x86_1h -mem 40g -cores "10+1" ./run_all_unit_tests.sh 11.6 /dccstor/fuse_med_ml/cicd/envs/
+                jbsub -wait -out ccc_log.txt -queue x86_1h -mem 40g -cores "10+1" -require 'v100' ./run_all_unit_tests.sh 11.6 /dccstor/fuse_med_ml/cicd/envs/
                 echo "------ printing ccc_log.txt -----"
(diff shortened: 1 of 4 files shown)


run_all_unit_tests.sh
```diff
@@ -30,7 +30,7 @@ create_env() {
         requirements+=$(cat examples/requirements.txt)
     fi
 
-    PYTHON_VER=3.7
+    PYTHON_VER=3.10
```
mosheraboh, line 33: Now that we see that it pass - you can change it to 3.8

Jenkinsfile
```diff
@@ -13,7 +13,7 @@ pipeline {
             steps {
                 sh '''
                 rm -rf test-reports/TEST-*.xml
-                jbsub -wait -out ccc_log.txt -queue x86_1h -mem 40g -cores "4+1" ./run_all_unit_tests.sh 11.6 /dccstor/fuse_med_ml/cicd/envs/
+                jbsub -wait -out ccc_log.txt -queue x86_1h -mem 40g -cores "10+1" ./run_all_unit_tests.sh 11.6 /dccstor/fuse_med_ml/cicd/envs/
```
mosheraboh, line 16: We though maybe to add also:
-require v100

mosheraboh (approved):
Great, thanks!
Can you please also update the readme file and remove "(3.7 is recommended)"?

Jenkinsfile
SagiPolaczek: Correct.
I accidentally added `-require *a100*` to ISIC's PR, so I'm updating both now.

mosheraboh (approved):
Looks good!

PR #184, opened 2022-09-29
multimodal modifications

mosheraboh:


fuse/data/datasets/sample_caching_audit.py
```diff
@@ -101,4 +101,5 @@ def audit(self, cached_sample, fresh_sample):
                 "Gladly, the Audit feature caught this stale cache state! :)\n"
                 f"sample id in which this staleness was caught: {get_sample_id(fresh_sample)}\n"
                 'NOTE: if small changes between the saved cached and the live-loaded/processed sample are ok for your use case, you can set a tolerance epsilon like this: audit_diff_kwargs={"math_epsilon":1e-9}'
+                f"diff = {diff}"
```
SagiPolaczek, line 104: Nice, does `diff` come out in a readable format?

fuse/data/ops/ops_common.py
```diff
@@ -381,6 +381,22 @@ def __call__(self, sample_dict: NDict, keep_keypaths: List[str]) -> Union[None,
         return sample_dict
 
 
+class OpDeleteKeypaths(OpBase):
+    """
+    Use this op to remove keypaths from the sample
+    A case where this is useful is if you want to limit the amount of data that gets transferred by multiprocessing by DataLoader workers.
+    """
+
+    def __init__(self, **kwargs):
```
SagiPolaczek, line 390: I think you can delete the `__init__` function and the logic remain the same. If so, I guess it's preferred. what do you think?

fuseimg/data/ops/color.py
```diff
@@ -63,8 +62,7 @@ def __init__(self, **kwargs):
 
     def __call__(self, sample_dict: NDict, key: str):
         img = sample_dict[key]
-        img = skimage.img_as_float(img)
-        sample_dict[key] = img
+        sample_dict[key] = img / img.max()
```
SagiPolaczek, line 65: Cool I remember we talked about changing that :) 
Shouldn't we also subtract the min value? When the minimum value isn't zero the range won't be exactly [0, 1]

SagiPolaczek (approved):
I added some comments inline, Thanks 😄 !

fuse/data/datasets/sample_caching_audit.py
mosheraboh: Yes, it's informative, but long

fuseimg/data/ops/color.py
mosheraboh: Good catch! - I copied it from ukbb branch without checking.

Issue #192, opened 2022-10-12
Pinned protobuf version exposes CVE-2022-1941


vagenas:
**Describe the bug**\
Protobuf dependency [currently pinned to 3.20.1](https://github.com/BiomedSciAI/fuse-med-ml/blob/9e46921852b16447e1beb9bb90edf447e514e42a/fuse/requirements.txt#L3) is exposing "high"-score vulnerability [CVE-2022-1941](https://nvd.nist.gov/vuln/detail/CVE-2022-1941).

**FuseMedML version**\
0.2.5 

**Python version**\
3.8.1

**To reproduce**\
Check the [currently pinned version](https://github.com/BiomedSciAI/fuse-med-ml/blob/9e46921852b16447e1beb9bb90edf447e514e42a/fuse/requirements.txt#L3).

**Expected behavior**\
A secure version of the dependency should be used.

PR #193, opened 2022-10-12
fix(#192): relax protobuf constraint

vagenas:
This is targeting #192.


mosheraboh:
Thanks @vagenas , you can go ahead and merge it.
@SagiPolaczek, I think we can remove the protobuf entirely - please remove it in a new PR if possible.

mosheraboh (approved):

PR #194, opened 2022-10-13
added silent mode to evaluator

smartdanny:


fuse/eval/evaluator.py
```diff
@@ -54,6 +54,7 @@ def eval(
         id_key: str = "id",
         batch_size: Optional[int] = None,
         output_dir: Optional[str] = None,
+        silent: Optional[str] = True
```
mosheraboh, line 57: The type should be bool

fuse/eval/evaluator.py
```diff
@@ -79,8 +80,10 @@ def eval(
                                               data must be an iterator of samples.
                                               A batch will be automatically created from batch_size samples
         :param output_dir: Optional - dump results to directory
+        :param silent: Optional - print results if true
```
mosheraboh, line 83: Remove "optoinal"

mosheraboh (changes requested):
Looks good @smartdanny, thanks!
See minor comment inline.
We enforce black code format. See #130 for instructions.

mosheraboh (approved):
Looks good.

Issue #215, opened 2022-11-21
Enable classification metrics to use logits and class indicators


afoncubierta:
The current implementation of the classification metrics expects class labels as integers. The user might want to use a model that outputs logits to estimate the class. 

In this case, it would be possible to spare the user from tuning the model, by enabling the metric computation to use logits or class scores directly. This can be easily done by using type_of_target function from sklearn and if the result is some sort of `continous` value, apply an argmax function.

An alternative is that the user will implement the argmax or even explore the operating point of the model, but the proposed solution would be the default in case the user takes no action.

PR #216, opened 2022-11-21
feat: added support for logits in some classification metrics #215

afoncubierta:
Addresses issue #215

### DIFF
--- a/fuse/eval/metrics/libs/classification.py
+++ b/fuse/eval/metrics/libs/classification.py
@@ -42,7 +42,7 @@ class MetricsLibClass:
         Compute auc roc (Receiver operating characteristic) score using sklearn (one vs rest)
         :param pred: prediction array per sample. Each element shape [num_classes]
         :param target: target per sample. Each element is an integer in range [0 - num_classes)
-        :param sample_weight: Optional - weight per sample for a weighted auc. Each element is  float in range [0-1] 
+        :param sample_weight: Optional - weight per sample for a weighted auc. Each element is  float in range [0-1]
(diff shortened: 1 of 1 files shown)


fuse/eval/metrics/libs/classification.py
```diff
@@ -138,14 +140,20 @@ def accuracy(
     ):
         """
         Compute accuracy score
-        :param pred: class prediction. Each element is an integer in range [0 - num_classes)
-        :param target: the target class. Each element is an integer in range [0 - num_classes)
+        :param pred: class prediction. Each element is an integer in range [0 - num_classes) or a float in range [0-1]
```
mosheraboh, line 143: It's not a float but a numpy array (dtype=float) - score per class, right?
Can you please also mention that in such a case argmax will be applied?

fuse/eval/metrics/libs/classification.py
```diff
@@ -138,14 +140,20 @@ def accuracy(
...
+        :param target: the target class. Each element is an integer in range [0 - num_classes) or its one hot encoded version
         :param sample_weight: Optional - weight per sample for a weighted score. Each element is  float in range [0-1]
         :return: accuracy score
         """
 
         pred = np.array(pred)
         target = np.array(target)
+
+        if type_of_target(pred) in ("continuous", "continuous-multioutput"):
```
mosheraboh, line 152: Great

mosheraboh (changes requested):
Thanks! Looks good.
Minor comment about the documentation inline.

mosheraboh (approved):
Thanks!

mosheraboh (approved):

Issue #217, opened 2022-11-21
Make BackboneResnet3D backward compatible


agiova:
**Describe the bug**\
BackboneResnet3D expects the instance variable `self.pool` which was not present in models trained and stored with fuse-med-ml<0.2.8
See: https://github.com/BiomedSciAI/fuse-med-ml/blob/master/fuse/dl/models/backbones/backbone_resnet_3d.py#L83

**FuseMedML version**\
FuseMedML 0.2.8 used.

**Python version**\
3.8.15

**To reproduce**\
Run forward pass with a model created with fuse-med-ml<0.2.8.

**Expected behavior**\
Models generated with previous versions should run.

agiova:
Thank you @mosheraboh for merging!

PR #218, opened 2022-11-21
adding check for class attribute, fixes issue #217

agiova:


fuse/dl/models/backbones/backbone_resnet_3d.py
```diff
@@ -80,7 +80,7 @@ def features(self, x: Tensor) -> Any:
         x = self.layer2(x)
         x = self.layer3(x)
         x = self.layer4(x)
-        if self.pool:
+        if "pool" in self.__dict__.keys() and self.pool:
```
mosheraboh, line 83:
```suggestion
        if hasattr(self, "pool") and self.pool:
```

mosheraboh (commented):
Thanks!
One suggestion inline.

mosheraboh (approved):

Issue #219, opened 2022-11-22
Support batching variable size tensors using nested tensors


mosheraboh:
**Is your feature request related to a problem? Please describe.**
Support batching variable size tensors using nested tensors (https://pytorch.org/tutorials/prototype/nestedtensor.html)
To avoid padding and improve the running time.

**Describe the solution you'd like**
Add such an option in CollateDefault as an alternative to CollateDefault.pad_all_tensors_to_same_size.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

SagiPolaczek:
Note that currently there is a user warning (torch 1.13.0):
```
UserWarning: The PyTorch API of nested tensors is in prototype stage and will change in the near future.
```

YoelShoshan:
if/when it works well it could be a very nice GRAM saver

PR #220, opened 2022-11-22
Add OpToNestedTensor (Prototype, for ref)

SagiPolaczek:
(NOT FOR MERGE)
Hey All,

Following @YoelShoshan's proposal and #219 I implemented some sort of prototype for supporting Nested Tensors.
Note that it requires latest PyTorch version (and it will to do so, since they still working on it). Current version is 1.13.0.

Example of how I used it in a data pipeline (DTI task, so the sequence are differ in size between samples!):

```python
        dynamic_pipeline = [
            # read data
            (OpReadDataframe(data=df, rename_columns=rename_columns, key_column=None), dict()),

            # dummy encoding (ASCII values) and conversion to tensors
            (OpDummyStringToNumbers(), dict(key_in="data.drug.smiles", key_out="data.drug.encoding")),
            (OpDummyStringToNumbers(), dict(key_in="data.target.sequence", key_out="data.target.encoding")),
            (OpToNumpy(), dict(key=["data.drug.encoding", "data.target.encoding"], dtype=np.float32)),
            (OpToTensor(), dict(key=["data.drug.encoding", "data.target.encoding"], dtype=torch.float)),

            # covert to nested tensor
            (OpToNestedTensor(), dict(keys_in=["data.drug.encoding", "data.target.encoding"], key_out="data.input.nested_tensor")),
        ]

```


SagiPolaczek:
for future ref:
The use of Nested Tensor should be consider to be in the collate phase, and not as a stand-alone op.

thanks @mosheraboh

SagiPolaczek:
Cleaning PRs

PR #222, opened 2022-11-30
Add OpTranspose (not critical)

smartdanny:
Had an issue when integrating perceiver architechture that all the implementations require channels to be the last input dim of the minibatch. This required me transposing after OpToTensor in isic dataloader.

This can also be done using OpLambda and passing torch.transpose just as easily, so its not a very necessary pull request. Only merge this if you want to @mosheraboh


mosheraboh (commented):
Thanks @smartdanny 
We do have ops OpCHWToHWC and OpHWCToCHW.
I think that it could be more general.
I thought about implementing op that moves from "channels first" to "channels last' format (and vice versa) even if it's not an image. What do you think?
@SagiPolaczek, we didn't implement it yet, right?

PR #224, opened 2022-12-04
mod for peptide design

mosheraboh:


fuse/data/datasets/dataset_default.py
```diff
@@ -185,7 +185,12 @@ def getitem(
             raise Exception("you must first call create()")
 
         # get sample id
-        if not isinstance(item, (int, np.integer)) or not self._explicit_sample_ids_mode:
+        if not self._explicit_sample_ids_mode:
```
SagiPolaczek, line 188: Not relevant for this PR, but I suggest we change the flag's name from `_explicit_...` to `_implicit_...` .

SagiPolaczek (approved):
LGTM 😄 

I'm curious about the context tho - I'll ask offline

fuse/utils/data/collate.py
```diff
@@ -34,13 +35,15 @@ def __init__(
         self,
         skip_keys: Sequence[str] = tuple(),
         raise_error_key_missing: bool = True,
+        missing_values: Sequence[str] = (None, "N/A")
```
SagiPolaczek, line 38: A documentation of this parameter is missing, tho I don't see a use of it so maybe it is supposed to be deleted?

fuse/data/utils/collates.py
```diff
@@ -80,10 +80,10 @@ def __call__(self, samples: List[Dict]) -> Dict:
 
             try:
                 # collect values into a list
-                collected_values, has_error = self._collect_values_to_list(samples, key)
+                collected_values, has_error, has_missing_values = self._collect_values_to_list(samples, key)
 
                 # batch values
-                self._batch_dispatch(batch_dict, samples, key, has_error, collected_values)
+                self._batch_dispatch(batch_dict, samples, key, has_error or has_missing_values, collected_values)
```
SagiPolaczek, line 86: The signature of `_batch_dispatch()` contains only `has_error` has a parameter. Should we add something to mention `has_missing_values` ?

SagiPolaczek: The signature is located in line 94

SagiPolaczek (commented):
Added two comments inline + black is missing :-)

SagiPolaczek (approved):
LGTM, Thanks! 😄

PR #226, opened 2022-12-08
Added more descriptive error messages and hints for common usage mistakes

YoelShoshan:
When providing a a pipeline description, there are common pitfalls.

It's better to identify those common mistakes and provide clear instructions on how to fix it.

In this PR I added it for two cases:

Case 1:
######
 You accidentally provide a class object instead of a class instance (very common mistake!)

Example exception message:

Exception: Error: expected an instance object, not a class object for <class 'fusedrug.data.protein.ops.aa_ops.OpToUpperCase'>
When creating a pipeline, such error can happen when you provide the following ops list description:
[SomeOp, {}]
instead of 
[SomeOp(), {}]

Case 2:
#####
you provided something that isn't a class or subclass of OpBase

Exception: Ops are expected to be instances of classes or subclasses of OpBase. The following op is not: 1234

In the future I think that we should also add:

1. Validation of all of this at the pipeline constructor level, since it's already known then
2. also validate the structure of the provided list of tuples with the pipeline op list


mosheraboh (approved):
Great, thanks!

PR #227, opened 2022-12-11
support arbitrary (not sequential and even not ints) balanced class values

mosheraboh:


SagiPolaczek (commented):
Thanks Moshiko !

I added few comments inline.
If I was mistaken in line 114 please see this review as approved

SagiPolaczek (approved):
Cool Thanks!

LGTM 😄

PR #233, opened 2022-12-21
Some changes from DTI task (Add doc' and some small features)

SagiPolaczek:
Hey!

In this PR I'm suggesting to add some changes I made while working on the drug target interaction (DTI) task - which is still in progress.

Changes:
1. Added more functionality to some Ops
2. Additional documentation & small fixes
3. Small modification to the NDict's `ndict. print_tree()` function so it will also print the key values (useful for printing nitely hydra's cfg objects!).  @egozi 

Example for  `.print_tree()`'s output:

BEFORE CHANGES:
```
>>> ndict = NDict()
>>> ndict["data.input.drug"] = "this_is_a_drug_seq"
>>> ndict["data.input.target"] = "this_is_a_target_seq"
>>> ndict.print_tree()
--- data
------ input
--------- drug
--------- target
```

AFTER CHANGES:
```
>>> ndict = NDict()
>>> ndict["data.input.drug"] = "this_is_a_drug_seq"
>>> ndict["data.input.target"] = "this_is_a_target_seq"
>>> 
>>> ndict.print_tree()
--- data
------ input
--------- drug
--------- target
>>> 
>>> ndict.print_tree(print_values=True)
--- data
------ input
--------- drug -> this_is_a_drug_seq
--------- target -> this_is_a_target_seq
```

### Example where it is useful to print `NDict`'s values:

Lets take hydra's cfg object in our pipeline.

We can define the following function:
```python
def print_cfg_tree(cfg):
    """
    beautifully print the config file
    """
    NDict(OmegaConf.to_object(cfg)).print_tree(print_values=True)
```
and then use it with:
```python
@hydra.main(config_path=".", config_name="config")
def main(cfg: DictConfig):
    cfg = hydra.utils.instantiate(cfg)
    print_cfg_tree(cfg)
    pass
```
To get the config printed in to the console as (please ignore the values themselves):
```
--- paths
------ root_dir -> ./_examples/dti
------ data_dir -> /dccstor/mm_hcls/usr/sagi/_data/BindingDB/
------ model_dir -> ./_examples/dti/mlp
------ cache_dir -> ./_examples/dti/cache_dir
------ infer_dir -> ./_examples/dti/mlp/infer_dir
------ eval_dir -> ./_examples/dti/mlp/eval_dir
------ ckpt_filename -> best_epoch.ckpt
------ infer_filename -> infer_file.gz
--- train
------ model -> mlp
------ num_workers -> 8
------ batch_size -> 32
------ lr -> 1e-05
------ weight_decay -> 0.001
------ num_epochs -> 1
------ accelerator -> gpu
------ num_devices -> 1
------ model_params
--------- mlp -> None
--------- cat -> None
--------- cat_custom -> None
--- data
------ drug_fixed_size -> 80
------ target_fixed_size -> 1000
--- logging
------ log_clear_ml -> 0
------ task_name -> mlp-
```

I used it just before moving to ClearML (which saves the hyperparams) and still using it in times I don't want to log into ClearML.
Other different use cases might be found.


fuse/dl/models/backbones/backbone_mlp.py
```diff
@@ -46,7 +46,9 @@ def __init__(
 
         self.mlp = nn.ModuleList(mlp_layers)
 
-    def forward(self, input_tensor: Dict) -> Dict:
+    def forward(self, input_tensor: torch.Tensor) -> torch.Tensor:
+        input_tensor = input_tensor.float()
```
SagiPolaczek, line 50: @mosheraboh 

I'm not sure that's the right practice. Feels fishy.
I had to use it in order to supports both the MLP and the Transformer with the same DTI dataset.

What do you think?

fuse/data/ops/ops_read.py
```diff
@@ -46,7 +46,7 @@ def __init__(
         :param columns_to_extract: list of columns to extract from dataframe. When None (default) all columns are extracted
         :param rename_columns: rename columns from dataframe, when None (default) column names are kept
         :param key_name: name of value in sample_dict which will be used as the key/index
-        :param key_column: name of the column which use as key/index
+        :param key_column: name of the column which use as key/index. set to None to do nothing.
```
mosheraboh, line 49: In case of None, the original dataframe index will be used to extract the values for a single sample

fuse/dl/models/backbones/backbone_mlp.py
mosheraboh: Yes,
What was the dtype before you've changed it?
Let's check which code generated this tensor.

fuse/utils/ndict.py
```diff
@@ -293,23 +293,53 @@ def get_multi(self, keys: Optional[List[str]] = None) -> NDict:
...
+        keys = data_dict.keys()
+        level += 1
+        for key in keys:
+            if type(data_dict[key]) == dict:
                 print("---" * level, key)
-                NDict._print_tree_static(data_dict[key], level)
+                NDict._print_tree_static(data_dict[key], level, print_values=print_values)
+            else:
+                if print_values:
+                    print("---" * level, key, "->", data_dict[key])
```
mosheraboh, line 340: Printing values seem a bit too much in cases you've big tensors.
Was it useful for you?

mosheraboh (changes requested):
Looks good!
Few minors inline.

fuse/dl/models/backbones/backbone_mlp.py
SagiPolaczek: @mosheraboh 
I'm casting it to `int` explicitly in the dataset using:
```
(OpToTensor(), dict(key=["data.drug.encoding", "data.target.encoding"], dtype=torch.int32)),
```

Do we expect dataset to return `float` all the time? If so lets delete this change and I'll take care of it in my DTI pipeline 😃

fuse/utils/ndict.py
SagiPolaczek: Yes it was. I'll add an example to the PR description right now.

fuse/data/ops/ops_read.py
SagiPolaczek: My bad. I don't even recall it 😅 

fixed.

fuse/utils/ndict.py
SagiPolaczek: DONE 😄

fuse/dl/models/backbones/backbone_mlp.py
mosheraboh: The input to the model should be float.
Not sure if that's what you asked.
Anyway, I think we should remove this line.

mosheraboh (approved):
looks good, thanks!

PR #238, opened 2023-01-01
fix learning rate and added config options

liamhazan:


fuseimg/datasets/knight.py
```diff
@@ -187,6 +200,7 @@ def static_pipeline(data_path: str, resize_to: Tuple, test: bool = False) -> Pip
                 # loading data
                 (OpLoadImage(data_path), dict(key_in="data.input.img_path", key_out="data.input.img", format="nib")),
                 # fixed image normalization
+                # (OpNormalizeAgainstSelf(), dict(key="data.input.img")),
```
mosheraboh, line 203: If you are not using it, please delete it and also the definition above,

fuse/data/datasets/sample_caching_audit.py
```diff
@@ -102,4 +102,10 @@ def audit(self, cached_sample, fresh_sample):
                 f"sample id in which this staleness was caught: {get_sample_id(fresh_sample)}\n"
                 'NOTE: if small changes between the saved cached and the live-loaded/processed sample are ok for your use case, you can set a tolerance epsilon like this: audit_diff_kwargs={"math_epsilon":1e-9}'
                 f"diff = {diff}"
+                f"Error! During AUDIT found a mismatch between cached_sample and loaded sample.\n"
```
mosheraboh, line 105: Can you please define the message above and then do something like (to avoid duplication):
msg + diff + msg

examples/fuse_examples/imaging/classification/knight/baseline/config.yaml
```diff
@@ -1,21 +1,25 @@
...
+num_epochs : 100
+learning_rate : 0.0003
 imaging_dropout : 0.5
 fused_dropout : 0.5
 testing : False # for testing\continous integration
+reset_cache : False
 
 task_1:
   num_classes : 2
+  class_names: None
```
mosheraboh, line 16: To get None in your code from yaml, you should set it to null

examples/fuse_examples/imaging/classification/knight/baseline/config.yaml
```diff
@@ -1,21 +1,25 @@
 experiment_num : 0
-task_num : task_2 # task_1 or task_2
+task_num : task_1 # task_1 or task_2
 num_gpus : 1
 use_data : {"imaging": True, "clinical": True}
-batch_size : 2
-resize_to : [70, 256, 256]
-num_epochs : 150
-learning_rate : 0.0001
+batch_size : 6
```
mosheraboh, line 5: Are you using a much bigger GPU?
If yes, can add in comment here that for smaller GPUs, the batch size should be smaller?

mosheraboh (changes requested):
Looks good,
Please find few specific comments inline.

fuseimg/datasets/knight.py
liamhazan: why? its an option for the user to try it

mosheraboh (approved):
Thanks!

mosheraboh:
@liamhazan, please see that the test fuse_examples.tests.test_classification_knight.KnightTestTestCase.test_train.
There are two more failures that you can ignore.

mosheraboh (approved):

PR #239, opened 2023-01-02
CICD fix

mosheraboh:


mosheraboh:
@avihu111 , there is an issue with the webhook and the unittests are not triggered automatically.
I reverted your last PR, so all the tests will pass.
Please recreate your PR, either we missed one scenario or we need to fix also the tests.

PR #240, opened 2023-01-03
Simple MNIST Example

smartdanny:
This mnist example is intended to be a friendly/approachable introduction to the core elements of fuse. Which is mostly our batch dict, ready-to-go datasets, and loss functions. 

- Simplified Example
- Removed Eval/Test
- Added some defaults to LightningModuleDefault so that it can just be inherited from and user can set model and model dir themselves (looks cleaner imo)

Happy to add back some functionality if anyone wants.
@mosheraboh @avihu111 @SagiPolaczek


examples/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
```diff
@@ -0,0 +1,142 @@
...
+    "opt.lr": 1e-4,
+    "opt.weight_decay": 0.001,
+}
+
+
+class FuseLitLenet(LightningModuleDefault):
+    def __init__(self):
+        super().__init__()
+        # wrap basic torch model to automatically read inputs from batch_dict and save its outputs to batch_dict
+        self._model = ModelWrapSeqToDict(
```
mosheraboh, line 75: you should pass all those objects to super().__init__() instead of overriding the internal attributes, It's cleaner.

examples/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
```diff
@@ -0,0 +1,142 @@
...
+        }
+        self._train_metrics = OrderedDict(
+            [
+                ("operation_point", MetricApplyThresholds(pred="model.output.classification")),  # will apply argmax
+                ("accuracy", MetricAccuracy(pred="results:metrics.operation_point.cls_pred", target="data.label")),
+            ]
+        )
+        self._validation_metrics = copy.deepcopy(self._train_metrics)
+        optimizer = optim.Adam(
+            self._model.parameters(), lr=train_params["opt.lr"], weight_decay=train_params["opt.weight_decay"]
```
mosheraboh, line 94: Don't read arguments directly from a global variable. Instead, get it as an argument in __init__

fuse/dl/lightning/pl_module.py
```diff
@@ -30,8 +31,8 @@ class LightningModuleDefault(pl.LightningModule):
 
     def __init__(
         self,
-        model_dir: str,
-        model: torch.nn.Module,
+        model_dir: str = None,
```
mosheraboh, line 34: Why?
model_dir and model are must also in your case.

examples/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+        lr_scheduler = optim.lr_scheduler.ReduceLROnPlateau(optimizer)
+        lr_sch_config = dict(scheduler=lr_scheduler, monitor="validation.losses.total_loss")
+        self._optimizers_and_lr_schs = dict(optimizer=optimizer, lr_scheduler=lr_sch_config)
+
+
+def run_train(paths: dict, train_params: dict):
+    train_dataset = MNIST.dataset(PATHS["cache_dir"], train=True)
```
mosheraboh, line 102: use paths and not the global variable.

examples/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
```diff
@@ -0,0 +1,142 @@
...
+
+ROOT = "_examples/mnist"  # TODO: fill path here
+model_dir = os.path.join(ROOT, "model_dir")
+PATHS = {
+    "model_dir": model_dir,
+    "cache_dir": os.path.join(ROOT, "cache_dir"),
+    "inference_dir": os.path.join(model_dir, "infer_dir"),
+    "eval_dir": os.path.join(model_dir, "eval_dir"),
+}
+train_params = {
```
mosheraboh, line 57: global var  -> change to upper case.

mosheraboh (changes requested):
Thanks @smartdanny.
The example looks good!
Do you find it a better example to start with than the one we currently have? Perhaps we should replace the one we have with this one.
General comments:
* Please also add inline comments (if we are trying to improve the basic examples :) )
* Create a unittest - just to check that nothing crashes.

examples/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
```diff
@@ -0,0 +1,164 @@
...
+from fuseimg.datasets.mnist import MNIST
+from examples.fuse_examples.imaging.classification.mnist import lenet
+
+## Paths and Hyperparameters ############################################
+ROOT = "_examples/mnist"  # TODO: fill path here
+MODEL_DIR = os.path.join(ROOT, "model_dir")
+PATHS = {
+    "model_dir": MODEL_DIR,
+    "cache_dir": os.path.join(ROOT, "cache_dir"),
+    "inference_dir": os.path.join(MODEL_DIR, "infer_dir"),
```
mosheraboh, line 56: You are not using inference dir and eval dir, let's delete it.

examples/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
```diff
@@ -0,0 +1,164 @@
...
+    "trainer.accelerator": "gpu",
+    "trainer.strategy": "dp",
+    "trainer.ckpt_path": None,
+    "opt.lr": 1e-4,
+    "opt.weight_decay": 0.001,
+}
+
+## Model Definition #####################################################
+class FuseLitLenet(LightningModuleDefault):
+    def __init__(self, paths, train_params):
```
mosheraboh, line 74: Let's be more precise here and specify model_dir, lr and weight_decay only.
In addition, please add type annotation and doc for the arguments.

fuse/dl/lightning/pl_module.py
```diff
@@ -15,6 +15,7 @@
 
 Created on June 30, 2021
 """
+from typing import Sequence, Dict
```
mosheraboh, line 18: Why? did you get an error without it?

mosheraboh (commented):
Looks good.
We talked about replacing the current example altogether with this example.
What's your take on this option?

fuse/dl/lightning/pl_module.py
smartdanny: Just forgot to remove :) its gone

examples/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
smartdanny: done :)

examples/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
smartdanny: done :)

smartdanny:
I think the added features like inference and evaluation are important to be shown somewhere, though not sure if mnist is the place. It depends on how expensive/annoying it is for us to "maintain" multiple training scripts for mnist and all of our other examples. If it costs us nothing, then I say we should just leave them. But if its even a little time consuming for us to maintain them, then I say at least for mnist we should get rid of the more complex scripts. Leave them for the more complex examples/datasets like knight and isic.

smartdanny:
@YoelShoshan I guess you are referring to the way that the model is constructed in the custom mnist example? https://github.com/BiomedSciAI/fuse-med-ml/blob/master/examples/fuse_examples/imaging/classification/mnist/run_mnist_custom_pl_imp.py

If so we can just replace the other mnist and leave this custom one.

SagiPolaczek (approved):
Thanks, LGTM!

Two general comments:
1. We can implement and use some `MNISTDataModule` which will reduce even more the runner's lines count. But it's not necessary of course.
2. I think it worth having the sanity check that look at the evaluation score, since it easy and fast to do so with the MNIST example it's useful to have it in the CICD.
I can open an issue regarding that once this PR will be merged.

PR #244, opened 2023-01-08
support not specifying number of channels in ModelMultiHead

alex-golts:


fuse/dl/models/model_multihead.py
```diff
@@ -33,8 +33,8 @@ def __init__(
         self,
         backbone: torch.nn.Module,
         heads: Sequence[torch.nn.Module],
-        conv_inputs: Tuple[Tuple[str, int], ...] = None,
-        backbone_args: Tuple[Tuple[str, int], ...] = None,
+        conv_inputs: Tuple[Tuple[str, int], str] = None,
```
mosheraboh, line 36: It should be:
Union[Sequence[str], Sequence[Tuple[str, int]]

fuse/dl/models/model_multihead.py
```diff
@@ -64,6 +64,10 @@ def __init__(
         self.add_module("heads", self.heads)
 
     def forward(self, batch_dict: NDict) -> Dict:
+        if isinstance(self.conv_inputs, str):  # no number of input channels specified
```
mosheraboh, line 67: Few comments:
* self.conv_inputs is always a sequence (or None).
   You should check about the type of an element in this sequence instead. 
* To keep the code simple let's convert the sequence of a tuple to a sequence of str instead (keep only the keys)
* Let's move this code to the constructor - to do it only once

mosheraboh (changes requested):
Thanks, @alex-golts!
 I've added two comments inline.
 Let me know if it doesn't make sense to you.

fuse/dl/models/model_multihead.py
alex-golts: or `Union[Tuple[str], Tuple[Tuple[str, int]]`, no? if we want it to precisely extend the current version

alex-golts: oh actually I see why not, because we then want to reassign it and tuple is immutable

fuse/dl/models/model_multihead.py
alex-golts: I addressed most. can you clarify what you mean in the 2nd bullet?

fuse/dl/models/model_multihead.py
```diff
@@ -54,6 +54,12 @@ def __init__(
             raise Exception(
                 "Neither conv_inputs nor backbone_args are None. One must be set (conv_inputs soon to be deprecated)"
             )
+        # we don't use the number of input channels.
+        # it's just kept for backward compatibility
+        if isinstance(conv_inputs[0], Tuple):
```
mosheraboh, line 59: conv_inputs (and also backbone_args) might be None (and maybe also empty?).
What do you think about:
```
if conv_inputs is not None:
    self.conv_inputs = [conv_input[0] for conv_input in self.conv_inputs if isinstance(conv_input, Tuple) else conv_input]
```

alex-golts: sounds good. just the syntax should be:
`[inp[0]  if isinstance(inp, Tuple) else inp for inp in self.conv_inputs]`

mosheraboh:
@alex-golts. this one doesn't wait for me, right?

alex-golts:
> @alex-golts. this one doesn't wait for me, right?

I think there was an issue with the CICD that prevented me from seeing where it fails, but now it looks like I can see what causes the error, so I will take a look

fuse/dl/models/model_multihead.py
```diff
@@ -54,6 +54,12 @@ def __init__(
...
                 "Neither conv_inputs nor backbone_args are None. One must be set (conv_inputs soon to be deprecated)"
             )
+        # we don't use the number of input channels.
+        # it's just kept for backward compatibility
+        if conv_inputs is not None:
+            self.conv_inputs = [inp[0] if isinstance(inp, Tuple) else inp for inp in conv_inputs]
+        if backbone_args is not None:
+            self.backbone_args = [inp[0] if isinstance(inp, Tuple) else inp for inp in backbone_args]
 
         self.conv_inputs = conv_inputs
```
mosheraboh, line 64: @alex-golts,

Here is your issue, you override here the conversion you did in lines 57-62

alex-golts: thanks, missed it!
still not sure how come when I ran the tests locally they passed (in the same branch), but never mind I guess :-)

mosheraboh (approved):

PR #246, opened 2023-01-10
add down sample option to BatchSamplerDefault

mosheraboh:


fuse/dl/models/heads/heads_1D.py
```diff
@@ -105,9 +105,6 @@ def forward(self, batch_dict: Dict) -> Dict:
             batch_dict["model.output." + self.head_name] = prediction
         else:
             logits = self.head_module(global_features)  # --> res.shape = [batch_size, 2, 1, 1]
-            if len(logits.shape) > 2:
-                logits = logits.squeeze(dim=3)  # --> res.shape = [batch_size, 2, 1]
```
SagiPolaczek, line 109: Why? It happens that we don't get to that scenario?

fuse/data/utils/samplers.py
```diff
@@ -61,7 +61,9 @@ def __init__(
                                         In mode 'exact' the weights should be integers that sums up to batch size.
                                         In mode 'approx' the weights should be floats that sums up to ~1
                                         If not specified, equal number of samples from each class will be used.
-        :param num_batches: optional - if set will force num_batches, otherwise num_batches will be automatically to go over each sample at least once (exactly or approximately).
+        :param num_batches: optional - if set to an int will force num_batches,
```
SagiPolaczek, line 64: Cool!
worth adding to the function signature that `str` is also a valid type for `num_batches`

SagiPolaczek (approved):
Thanks! LGTM 

One small comment and one question inline :)

fuse/data/utils/samplers.py
mosheraboh: Thanks.
I will add it in my next PR though.

fuse/dl/models/heads/heads_1D.py
mosheraboh: It's a copy-paste from 3D head.

PR #251, opened 2023-01-10
Cross Attention Transformer

SagiPolaczek:
🚧 under construction 🚧


fuse/dl/models/backbones/backbone_transformer.py
```diff
@@ -45,3 +47,119 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         x = self.dropout(x)
         x = self.transformer(x)
         return x
+
+
+class CrossAttentionTransformer(nn.Module):
+    """
+    CrossAttentionTransformer backbone model based on x-transformers library.
+
```
avihu111, line 55: suggestion: 
maybe add some link to a paper/blogpost on the component?

SagiPolaczek: @avihu111 
Good idea , thanks :)

fuse/dl/models/backbones/backbone_transformer.py
```diff
@@ -36,12 +38,116 @@ def __init__(
...
         x = torch.cat((cls_tokens, x), dim=1)
         x += self.pos_embedding
         x = self.dropout(x)
         x = self.transformer(x)
         return x
+
+
+class CrossAttentionTransformerEncoder(nn.Module):
+    """
+    CrossAttentionTransformerEncoder backbone model based on x-transformers library.
```
mosheraboh, line 54: Can you describe here the arch?
What is the forward pass of seqa and seqb and where do you apply the cross attention

fuse/dl/models/backbones/backbone_transformer.py
```diff
@@ -36,12 +38,116 @@ def __init__(
...
+        super().__init__()
+
+        if output_dim is None:
+            output_dim = emb_dim
+
+        assert context in ["seq_a", "seq_b", "both"]
+        self._context = context
+
+        # init sequences' encoders
+        self.enc_a = TransformerWrapper(
```
mosheraboh, line 106: Are there any additional arguments to TransformerWrapper or Encoder that it we should expose?

fuse/dl/models/backbones/backbone_transformer.py
mosheraboh: You should also specify TransformerWrapper and Encoder so the user can understand things like how the tokens converted to embedding

mosheraboh (approved):
Looks good. 
Few comments about documentation

fuse/dl/models/backbones/backbone_transformer.py
SagiPolaczek: Yes. actually there are a lot more.
I will add them as kwargs 😄

fuse/dl/models/backbones/backbone_transformer.py
```diff
@@ -36,12 +38,146 @@ def __init__(
...
+            num_tokens=num_tokens_a,
+            max_seq_len=max_seq_len_a,
+            **kwargs_wrapper_a,
```
mosheraboh, line 137: You marked kwargs_wrapper_a as optional - but it seems like you'll get an error if you don't set it.

SagiPolaczek: nice catch! missed that, thanks!

mosheraboh (approved):
Cool
You can go ahead and merge.

PR #259, opened 2023-01-22
allow non-int sample ids in DatasetDefault

alex-golts:


fuse/data/datasets/dataset_default.py
```diff
@@ -187,8 +187,9 @@ def getitem(
         # get sample id
         if not self._explicit_sample_ids_mode:
             sample_id = item
-            if sample_id >= self._final_sample_ids:
-                raise IndexError
+            if isinstance(sample_id, (int, np.integer)):  # allow using non int sample_ids
```
mosheraboh, line 190: So it means that works?
If yes, for the official fix - I want to support it more explicitly.
In the constructor, 
We will support setting sample_ids to None.
If it will set to None - it means that Dataset has no control over the used sample ids, and therefore iter() and len() will raise an error.

mosheraboh (changes requested):
Thanks Alex!
See inline.
We can discuss it if it's clear or if you don't agree.

fuse/data/datasets/dataset_default.py
alex-golts: yes it worked. 
I made some changes, let me know if it's better

fuse/data/datasets/dataset_default.py
```diff
@@ -38,17 +38,21 @@
...
         dynamic_pipeline: Optional[PipelineDefault] = None,
         cacher: Optional[SamplesCacher] = None,
         allow_uncached_sample_morphing: bool = False,
     ):
         """
-        :param sample_ids:      list of sample_ids included in dataset.
-            Optionally, you can provide an integer that describes only the size of the dataset. This is useful in massive datasets
-             (for example 100M samples). In such case, multiple functionalities will not be supported, mainly -
-              cacher, allow_uncached_sample_morphing and get_all_sample_ids
+        :param sample_ids: list of sample_ids included in dataset. Or:
```
mosheraboh, line 48: Nice!

fuse/data/datasets/dataset_default.py
```diff
@@ -67,9 +71,11 @@ def __init__(
                 )
             if cacher is not None:
                 raise Exception("providing a cacher is not allowed when providing sample_ids=an integer value")
-            self._explicit_sample_ids_mode = False
+            self._explicit_sample_ids_mode = "running_int"
```
mosheraboh, line 74: Remove the explicit from self._explicit_sample_ids_mode.

fuse/data/datasets/dataset_default.py
```diff
@@ -67,9 +71,11 @@ def __init__(
...
+        elif sample_ids is None:
+            self._explicit_sample_ids_mode = None
```
mosheraboh, line 76: let's give it a name: "external" / "user" / something else that sounds better

fuse/data/datasets/dataset_default.py
```diff
@@ -185,10 +191,11 @@ def getitem(
             raise Exception("you must first call create()")
 
         # get sample id
-        if not self._explicit_sample_ids_mode:
+        if self._explicit_sample_ids_mode != "explicit":
             sample_id = item
-            if sample_id >= self._final_sample_ids:
-                raise IndexError
+            if isinstance(sample_id, (int, np.integer)):  # allow using non int sample_ids
```
mosheraboh, line 196: if self._explicit_sample_ids_mode == "running_int":

mosheraboh (approved):
Looks great!
Minors inline

mosheraboh (approved):

PR #260, opened 2023-01-22
not strict mode in ProjectPatchesTokenizer

avihu111:
add strict boolean argument.
when strict=False, it allows to patchify incompatible shapes.


fuse/dl/models/backbones/backbone_vit.py
```diff
@@ -17,13 +17,22 @@ class ProjectPatchesTokenizer(nn.Module):
...
 
-    def __init__(self, *, image_shape: Sequence[int], patch_shape: Sequence[int], channels: int, token_dim: int):
+    def __init__(
+        self,
+        *,
+        image_shape: Sequence[int],
+        patch_shape: Sequence[int],
+        channels: int,
+        token_dim: int,
+        strict: bool = True,
```
mosheraboh, line 27: Can you please specify in a comment what strict and what happens if it's False and what happens if the Image dimensions must be divisible by the patch size

mosheraboh (approved):
Thanks Avihu!
Approved, just one minor comment inline

SagiPolaczek (approved):
Thanks!
re-approved after requested changes were addressed.

PR #262, opened 2023-01-23
support shallow copy of pipelines

michalozeryflato:


michalozeryflato:
made this change to allow shallow copy & extended of dynamic pipeline  (adding augmentation during train)

fuse/data/pipelines/pipeline_default.py
```diff
@@ -54,6 +56,12 @@ def __init__(
             self._op_ids = op_ids
         self._verbose = verbose
 
+    def copy(self):
```
mosheraboh, line 59: Nice,
Can you please add in a comment that it is a shallow copy (both pipelines will still use the same instances of ops)?

mosheraboh (approved):
Nice
Minor comment inline

fuse/data/pipelines/pipeline_default.py
michalozeryflato: done

mosheraboh (approved):

PR #266, opened 2023-01-28
add props to dataset and pipelines + small fix to cacher

michalozeryflato:
1. I added these properties to allow easy access to operation objects in UKBB dataset's pipelines. 
In my case, I used it to retrieve the number of features from a "dataframe loading operation". After the dataset was created this operation stored the number of features in one of its internal field members.

2. small fix for cacher - moved the writing of the pipeline description after the creation of the directory

### DIFF
--- a/fuse/data/datasets/caching/samples_cacher.py
+++ b/fuse/data/datasets/caching/samples_cacher.py
@@ -207,14 +207,16 @@ class SamplesCacher:
             orig_sid_to_final[initial_sample_id] = output_sample_ids
 
         write_dir = self._get_write_dir()
+
+        set_info_dir = os.path.join(write_dir, "full_sets_info")
+        os.makedirs(set_info_dir, exist_ok=True)
+
         pipeline_desc_file = os.path.join(write_dir, f"pipeline_{self._pipeline_desc_hash}_desc.txt")
         if not os.path.exists(pipeline_desc_file):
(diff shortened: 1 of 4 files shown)


fuse/data/datasets/caching/samples_cacher.py
```diff
@@ -207,14 +207,16 @@ def cache_samples(self, orig_sample_ids: List[Any]) -> List[Tuple[str, Union[Non
             orig_sid_to_final[initial_sample_id] = output_sample_ids
 
         write_dir = self._get_write_dir()
+
+        set_info_dir = os.path.join(write_dir, "full_sets_info")
+        os.makedirs(set_info_dir, exist_ok=True)
+
         pipeline_desc_file = os.path.join(write_dir, f"pipeline_{self._pipeline_desc_hash}_desc.txt")
```
mosheraboh, line 214: pipeline_desc_file is not in set_info_dir.
What was your error?

mosheraboh (approved):
I'm not sure why you need the change in samples_cacher.py
But it seems ok to me in both ways,
The rest LGTM

fuse/data/datasets/caching/samples_cacher.py
michalozeryflato: makedirs create all the directories in the path to the root-dir.

PR #268, opened 2023-02-01
optimized train time for a use case of small samples and large batch

mosheraboh:
…size

### DIFF
--- a/fuse/data/utils/collates.py
+++ b/fuse/data/utils/collates.py
@@ -68,9 +68,11 @@ class CollateDefault(CollateToBatchList):
         batch_dict = NDict()
 
         # collect all keys
-        keys = self._collect_all_keys(samples)
         if self._keep_keys:
-            keys = [k for k in keys if k in self._keep_keys]
+            keys = self._keep_keys
+        else:
+            keys = self._collect_all_keys(samples)
+
         # collect values
         for key in keys:
 
--- a/fuse/utils/data/collate.py
+++ b/fuse/utils/data/collate.py
@@ -24,8 +24,6 @@ from fuse.utils import NDict
 import torch
 import numpy as np
 
-import math
-
 
 class CollateToBatchList(Callable):
     """
@@ -36,7 +34,6 @@ class CollateToBatchList(Callable):
         self,
         skip_keys: Sequence[str] = tuple(),
         raise_error_key_missing: bool = True,
-        missing_values: Sequence[str] = (None, "N/A"),
     ):
         """
         :param skip_keys: do not collect the listed keys
@@ -44,7 +41,6 @@ class CollateToBatchList(Callable):
         """
         self._skip_keys = skip_keys
         self._raise_error_key_missing = raise_error_key_missing
-        self._missing_values = missing_values
 
     def __call__(self, samples: List[Dict]) -> Dict:
         """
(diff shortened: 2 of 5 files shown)


fuse/utils/data/collate.py
```diff
@@ -98,22 +94,15 @@ def _collect_values_to_list(self, samples: List[str], key: str) -> Tuple[List, b
...
+            try:
+                value = sample[key]
+            except:
                 has_error = True
                 has_missing_values = True
                 if self._raise_error_key_missing:
                     raise Exception(f"Error: key {key} does not exist in sample {index}: {sample}")
                 else:
                     value = None
-            else:
```
mosheraboh, line 109: removing to optimize the running time.
Detecting NaNs and more could move to an optional op (in the data pipeline)

fuse/utils/ndict.py
```diff
@@ -228,7 +246,7 @@ def pop(self, key: str) -> Any:
         del self[key]
         return res
 
-    def indices(self, indices: Union[torch.Tensor, numpy.ndarray]) -> dict:
+    def indices(self, indices: Optional[numpy.ndarray]) -> dict:
```
mosheraboh, line 249: self review: will remove the optional here

fuse/utils/data/collate.py
```diff
@@ -124,49 +113,50 @@ def uncollate(batch: Dict) -> List[Dict]:
...
+    if "data.sample_id" in batch:
         batch_size = len(batch["data.sample_id"])
     else:
         batch_size = None
 
-    if batch_size is None:
+        keys = batch.keys()
+
         for key in keys:
             if isinstance(batch[key], torch.Tensor):
```
SagiPolaczek, line 132: Why there are two different loops? one for each case - `torch.Tensor`, `(np.ndarray, list)`
If I'm not missing something we can check for the two cases in the same loop.

fuse/utils/ndict.py
```diff
@@ -111,23 +111,41 @@ def flatten(self) -> dict:
...
-            if isinstance(self._stored[key], MutableMapping):
-                all_sub_keys = NDict(self[key]).flatten()
-                keys_to_add = {f"{key}.{sub_key}": all_sub_keys[sub_key] for sub_key in all_sub_keys}
-                all_keys.update(keys_to_add)
-            else:
-                all_keys[key] = self._stored[key]
-
-        return all_keys
+    @staticmethod
+    def _flatten_static(item: Union[dict, Any], prefix: str, flat_dict: dict) -> None:
```
SagiPolaczek, line 119: COOL!

fuse/utils/ndict.py
```diff
@@ -111,23 +111,41 @@ def flatten(self) -> dict:
...
+                NDict._flatten_static(value, cur_prefix, flat_dict)
+        else:
+            flat_dict[prefix] = item
 
     def keypaths(self) -> List[str]:
         """
         returns a list of keypaths (i.e. "a.b.c.d") to all values in the nested dict
         """
-        return list(self.flatten().keys())
+        return NDict._keypaths_static(self._stored, None)
```
SagiPolaczek, line 134: Why not using the same paradigm as before? Just calling `flatten()`? 
The two static functions has a lot in common

SagiPolaczek (commented):
Thanks!!

I added few comment inline to consider :)

fuse/data/utils/collates.py
```diff
@@ -68,9 +68,11 @@ def __call__(self, samples: List[Dict]) -> Dict:
         batch_dict = NDict()
 
         # collect all keys
-        keys = self._collect_all_keys(samples)
         if self._keep_keys:
```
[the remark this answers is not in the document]
mosheraboh: Good catch Now we will throw an error in such a case. I will update the comment.

SagiPolaczek (approved):
Thanks again!
I added comments in a different review 😄 

(I used the vscode interface and because of the changes in the middle it didn't allow me to approve)

fuse/utils/data/collate.py
mosheraboh: Because, first I want to look for tensor (trust it more), and if I can't find one then my second choice is (np.ndarray, list)

fuse/utils/ndict.py
mosheraboh: Mostly cause I don't want the overhead of creating a dictionary and extracting the keys,

PR #269, opened 2023-02-02
add multi-label metric cladd and func

egozi:


fuse/eval/metrics/libs/classification.py
```diff
@@ -61,6 +61,44 @@ def auc_roc(
...
+        if average == "macro":
+            # compute the roc for each class (column in y_score/y_true) and average
+            n_class = y_score.shape[1]
+            all_auc = []
+            for i in range(n_class):
+                auc = metrics.roc_auc_score(
+                    y_score=y_score[:, i], y_true=y_true[:, i], sample_weight=sample_weight, max_fpr=max_fpr
+                )
+                all_auc.append(auc)
+
```
mosheraboh, line 93: missing return of average of all_auc

fuse/eval/metrics/libs/classification.py
```diff
@@ -61,6 +61,44 @@ def auc_roc(
...
+        :return auc Receiver operating characteristic score
+        """
+        y_score = np.asarray(pred)
+        y_true = np.asarray(target)
+        if average == "macro":
+            # compute the roc for each class (column in y_score/y_true) and average
+            n_class = y_score.shape[1]
+            all_auc = []
+            for i in range(n_class):
+                auc = metrics.roc_auc_score(
```
mosheraboh, line 89: We should be careful here and ignore labels that don't have both positive class and negative class

fuse/eval/metrics/libs/classification.py
```diff
@@ -61,6 +61,44 @@ def auc_roc(
             y_score=y_score, y_true=np.asarray(target) == pos_class_index, sample_weight=sample_weight, max_fpr=max_fpr
         )
 
+    @staticmethod
+    def auc_roc_mult_label(
```
mosheraboh, line 65: multi_binary_label
Let's add the work binary next to multi_label to be more precise.

fuse/eval/metrics/libs/classification.py
```diff
@@ -61,6 +61,44 @@ def auc_roc(
...
+        pred: Sequence[Union[np.ndarray, float]],
+        target: Sequence[Union[np.ndarray, int]],
+        sample_weight: Optional[Sequence[Union[np.ndarray, float]]] = None,
+        average: str = "micro",
+        max_fpr: Optional[float] = None,
+    ) -> float:
+        """
+        Compute multi label auc roc (Receiver operating characteristic) score using sklearn
+        :param pred: prediction array per sample. Each element shape [num_classes]
```
mosheraboh, line 74: lets change num_classes -> num_labels

fuse/eval/metrics/libs/classification.py
```diff
@@ -61,6 +61,44 @@ def auc_roc(
...
+        :param target: target per sample. Each element shape [num_classes] with 0 and 1 only.
+        :param sample_weight: Optional - weight per sample for a weighted auc. Each element is  float in range [0-1]
+        :param pos_class_index: the class to compute the metrics in one vs rest manner - set to 1 in binary classification
```
mosheraboh, line 77: not relevant here

fuse/eval/metrics/classification/metrics_classification_common.py
```diff
@@ -145,6 +145,28 @@ def __init__(
         super().__init__(pred, target, metric_func=auc_roc, class_names=class_names, **kwargs)
 
 
+class MetricAUCROCmultLabel(MetricMultiClassDefault):
```
mosheraboh, line 148: MetricDefault instead of MetricMultiClassDefault

fuse/eval/metrics/classification/metrics_classification_common.py
```diff
@@ -145,6 +145,28 @@ def __init__(
...
+    """
+    Compute auc roc (Receiver operating characteristic) score using sklearn (one vs rest)
```
mosheraboh, line 150: Update the comment

fuse/eval/metrics/classification/metrics_classification_common.py
```diff
@@ -145,6 +145,28 @@ def __init__(
...
+    def __init__(
+        self,
+        pred: str,
+        target: str,
+        class_names: Optional[Sequence[str]] = None,
+        max_fpr: Optional[float] = None,
+        **kwargs,
+    ):
+        """
+        See MetricMultiClassDefault for the missing params
```
mosheraboh, line 162: Lets specify here explicitly what is the expected format of pred and target.

fuse/eval/metrics/classification/metrics_classification_common.py
```diff
@@ -145,6 +145,28 @@ def __init__(
...
+class MetricAUCROCmultLabel(MetricMultiClassDefault):
+    """
+    Compute auc roc (Receiver operating characteristic) score using sklearn (one vs rest)
+    """
+
+    def __init__(
+        self,
+        pred: str,
+        target: str,
+        class_names: Optional[Sequence[str]] = None,
```
mosheraboh, line 157: You are not using class_names, right?

fuse/eval/metrics/libs/classification.py
```diff
@@ -61,6 +61,44 @@ def auc_roc(
...
+            # compute the roc for each class (column in y_score/y_true) and average
+            n_class = y_score.shape[1]
+            all_auc = []
+            for i in range(n_class):
+                auc = metrics.roc_auc_score(
+                    y_score=y_score[:, i], y_true=y_true[:, i], sample_weight=sample_weight, max_fpr=max_fpr
+                )
+                all_auc.append(auc)
+
+        elif average == "micro":  # micro average
```
mosheraboh, line 94: Can you add a link that suggest this method?

mosheraboh (changes requested):
Thanks Amir, looks good.
Please see my comments inline.

Please also list it here: https://github.com/BiomedSciAI/fuse-med-ml/tree/master/fuse/eval#implemented-low-level-metrics

And please add example and a test: here (https://github.com/BiomedSciAI/fuse-med-ml/blob/master/fuse/eval/examples/examples.py) and here (https://github.com/BiomedSciAI/fuse-med-ml/blob/master/fuse/eval/tests/test_eval.py)

fuse/eval/metrics/libs/classification.py
egozi: it is still the number of classes. (num labels is the number of 1's In the vector)

fuse/eval/metrics/libs/classification.py
egozi: this is better in case there is no positive labels in one column. 
here the ref - 
In a multi-class classification setup with highly imbalanced classes, micro-averaging is preferable over macro-averaging. In such cases, one can alternatively use a weighted macro-averaging, not demoed here.
https://scikit-learn.org/stable/auto_examples/model_selection/plot_roc.html

fuse/eval/metrics/classification/metrics_classification_common.py
```diff
@@ -145,6 +145,28 @@ def __init__(
...
+        average: str = "micro",
+        **kwargs,
+    ):
+        """
+        See MetricMultiClassDefault for the missing params
+        :param max_fpr: float > 0 and <= 1, default=None
+                        If not ``None``, the standardized partial AUC over the range [0, max_fpr] is returned.
+        """
+        auc_roc = partial(MetricsLibClass.auc_roc_mult_binary_label, average=average, max_fpr=max_fpr)
+        super().__init__(pred, target, metric_func=auc_roc, **kwargs)
```
mosheraboh, line 167: To fix your problem, you specify the name of the arguments explicitly cause the order is different.
i.e. (pred=pred, target=target, ...) 
Please also change the name to MetricAUCROCMultLabel (capital M)

egozi: fixed

mosheraboh (approved):

PR #275, opened 2023-02-20
Fixes for the Hello World notebook

SagiPolaczek:
## Visit [here](https://colab.research.google.com/github/BiomedSciAI/fuse-med-ml/blob/sagi/hello_world_fix/fuse_examples/imaging/hello_world/hello_world.ipynb) to watch the updated notebook

So following #278 , I encountered [this annoying problem](https://stackoverflow.com/questions/57831187/need-to-restart-runtime-before-import-an-installed-package-in-colab) which makes us to restart the google colab notebook's runtime after the first Fuse's installation. Unfortunately I couldn't find a really good solution for that. The best workaround I came up with is to add a clear explanation for the FuseMedML installation 😵 . (**OPEN FOR SUGGESTIONS**)

### Changes:
1. Fixed prefix for model's outputs
2. Added a clear explanation for the FuseMedML installation on the google colab notebook.
3. **Improved UI**
4. Re-open the notebook's unit-test

Thanks @simona-rc for raising those issues 😄 

Update:
double checked on Jenkins:
Screenshot 2023-02-22 at 19 34 45

Update 2:
Fix typos in the installation instructions:
`[all+examples]` -> `[all,examples]`


mosheraboh (approved):
Great, thanks a lot also on the extra fixes!

PR #281, opened 2023-02-23
Pl model save + export the entire dataset (all keys) to dataframe

mosheraboh:


fuse/data/utils/export.py
```diff
@@ -61,9 +62,13 @@ def export_to_dataframe(
         # store in dataframe
         df = pds.DataFrame()
 
-        for key in all_keys:
-            values = [sample_dict[key] for sample_dict in data]
-            df[key] = values
+        if all_keys is None:
```
SagiPolaczek, line 65: nice catch

fuse/data/ops/op_base.py
```diff
@@ -107,11 +107,11 @@ def op_call(op: OpBase, sample_dict: NDict, op_id: str, **kwargs):
     try:
         if isinstance(op, OpReversibleBase):
             return op(sample_dict, op_id=op_id, **kwargs)
-        elif isinstance(op, OpBase):  # OpBase but not reversible
+        elif isinstance(op, OpBase) or hasattr(op, "__call__"):  # OpBase but not reversible
```
SagiPolaczek, line 110: Why? 🙂

SagiPolaczek (approved):
Awesome! LGTM 🚀 

Few questions / future suggestions:
1. Why `save_arguments = False` by default?
2. Do we have any unit-test for the save&load functionality? If not I think we should open an issue regarding that. What do you think?
3. Maybe in the future we can supply a wrapper for the checkpoint loading. In the default case (where the user use the default key names) we only need the "model_dir" parameter.
4. Curious, why did you do the following change?
```
elif isinstance(op, OpBase) or hasattr(op, "__call__"):
```
Thanks!

mosheraboh:
Thanks for the review @SagiPolaczek !
Even though we talked answering here for future reference:
1. the stored arguments are not useful in a common scenario (such as continue the training - cause you to run the exact same script anyway and recreate the arguments). model on the other hand is useful to load a model for inference.
2. Yes, we need unittest. I will open an issue. 
3. since it's three lines, I'm not sure.
4. I wanted to support the simplest building blocks in data pipeline (such as a function that gets sample_dict) and not to force all to inherit from OpBase.

PR #286, opened 2023-03-06
added ability to ignore kwargs in function string descriptor building logic, and solved an issue in run_multiprocessed in verbose=0 case

YoelShoshan:


mosheraboh (approved):
LGTM

PR #289, opened 2023-03-12
Make logging optional

smartdanny:


fuse/dl/lightning/pl_module.py
```diff
@@ -82,8 +82,12 @@ def __init__(
         """
         super().__init__(**kwargs)
 
+        if (save_arguments or save_model) and (model_dir is None):
```
mosheraboh, line 85: Are you avoiding from lightning logging? or maybe it's configurable in trainer?

fuse/dl/lightning/pl_module.py
```diff
@@ -32,8 +32,8 @@ class LightningModuleDefault(pl.LightningModule):
 
     def __init__(
         self,
-        model_dir: str,
         model: torch.nn.Module,
+        model_dir: str = None,
```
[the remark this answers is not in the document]
mosheraboh: Please also update the comments and change the type to Optional[str]

mosheraboh (changes requested):
Thanks @smartdanny !
See my comments inline and let me know if don't agree with any of them.

fuse/dl/lightning/pl_module.py
smartdanny: Yes, users can either only use lightning logging, or not log at all (which can be useful for debugging).

mosheraboh (approved):
LGTM

PR #291, opened 2023-03-16
few fixes to enable pl2 - and limit pl version until all tests will  …

mosheraboh:
…be fixed

The limitation will be removed by #292

### DIFF
--- a/fuse/dl/models/backbones/backbone_resnet.py
+++ b/fuse/dl/models/backbones/backbone_resnet.py
@@ -49,6 +49,9 @@ class BackboneResnet(ResNet):
         super().__init__(*init_parameters)
 
         # load pretrained parameters if required
+        if weights is not None and pretrained:
+            raise Exception("Use only one method to load pre-trained weights. Two were given!")
+
         if pretrained:
             print("Warning: not supported by new torchvision version - use weights instead")
             from torch.hub import load_state_dict_from_url
--- a/fuse/dl/models/backbones/backbone_resnet.py
+++ b/fuse/dl/models/backbones/backbone_resnet.py
@@ -29,7 +29,14 @@ class BackboneResnet(ResNet):
     2D ResNet backbone
     """
 
-    def __init__(self, *, pretrained: bool = False, weights: Optional[WeightsEnum] = None,in_channels: int = 3, name: str = "resnet18") -> None:
+    def __init__(
+        self,
+        *,
+        pretrained: bool = False,
+        weights: Optional[WeightsEnum] = None,
+        in_channels: int = 3,
+        name: str = "resnet18"
+    ) -> None:
         """
         Create 2D Resnet
         :param pretrained: reload imagenet weights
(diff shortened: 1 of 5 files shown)


smartdanny:
Looks good! with these changes I confirmed the mnist example works in this pr: https://github.com/BiomedSciAI/fuse-med-ml/pull/294

fuse/requirements.txt
```diff
@@ -8,7 +8,7 @@ matplotlib>=3.3.3
 scikit-learn>=0.23.2
 termcolor>=1.1.0
 pycocotools>=2.0.1
-pytorch_lightning>=1.6
+pytorch_lightning>=1.6,<2.0.0 # temp - need to make all tests pass with this version first
```
SagiPolaczek, line 11: not related to this PR, but generally speaking PL version should be higher to fully support DDP..

fuse/dl/models/backbones/backbone_resnet.py
```diff
@@ -47,9 +50,16 @@ def __init__(self, pretrained: bool = False, in_channels: int = 3, name: str = "
 
         # load pretrained parameters if required
         if pretrained:
+            print("Warning: not supported by new torchvision version - use weights instead")
+            from torch.hub import load_state_dict_from_url
```
SagiPolaczek, line 54: Do we want to keep support that option?

I suggest to use the `name` variable to pick the right wights in the new API ([API ref](https://pytorch.org/vision/stable/models.html) )

fuse/dl/models/backbones/backbone_resnet.py
```diff
@@ -47,9 +50,16 @@ def __init__(self, pretrained: bool = False, in_channels: int = 3, name: str = "
...
+            from torchvision.models.resnet import model_urls
+
             state_dict = load_state_dict_from_url(model_urls[name])
             self.load_state_dict(state_dict)
 
+        if weights is not None:
```
SagiPolaczek, line 60: I suggest to add the following to make sure the user don't supply the two arguments (because one overrides the other)

```python
if weights and pretrained :
  raise Exception("Use only one method to load pre-trained weights. Two were given!")
```

(or maybe another msg 😅 )

fuse/dl/models/backbones/backbone_resnet_3d.py
```diff
@@ -139,6 +137,10 @@ def __init__(
         self._pretrained = pretrained
 
         if pretrained:
+            print("Warning: not supported by new torchvision version")
```
SagiPolaczek, line 140: Also here :)

SagiPolaczek (commented):
Thanks!! Looks good!

Added few comments inline 😄

fuse/requirements.txt
mosheraboh: please add this request to #292

fuse/dl/models/backbones/backbone_resnet.py
mosheraboh: Not really, for backward compatibility.
I agree - just wanted the cicd to pass for now. will open an issue.

PR #294, opened 2023-03-16
Support for pl2 in simple mnist example

smartdanny:
The only thing I had to change was that auto_select_gpus is deprecated. Also, now you can use "auto" for strategy which is nice :)

### DIFF
--- a/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
+++ b/fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
@@ -63,7 +63,6 @@ TRAIN_PARAMS = {
     "trainer.num_epochs": 2,
     "trainer.num_devices": 1,
     "trainer.accelerator": "gpu",
-    "trainer.strategy": "dp",
     "trainer.ckpt_path": None,
     "opt.lr": 1e-4,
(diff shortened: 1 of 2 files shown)


fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
```diff
@@ -166,9 +165,8 @@ def run_train(paths: dict, train_params: dict):
         default_root_dir=paths["model_dir"],
         max_epochs=train_params["trainer.num_epochs"],
         accelerator=train_params["trainer.accelerator"],
-        strategy=train_params["trainer.strategy"],
+        strategy="auto",
```
mosheraboh, line 168: setting auto doesn't work for <2.0.0 versions.
Let's just not set it for now? (should work for both versions)

mosheraboh (commented):
Thanks!
One comment inline

fuse_examples/imaging/classification/mnist/simple_mnist_starter.py
smartdanny: Done :)

mosheraboh (approved):

PR #297, opened 2023-03-22
add postprocess to lostDefault

itaijj:


fuse/eval/metrics/libs/classification.py
```diff
@@ -76,10 +77,13 @@ def auc_roc_mult_binary_label(
         :param sample_weight: Optional - weight per sample for a weighted auc. Each element is  float in range [0-1]
         :param max_fpr: float > 0 and <= 1, default=None
                         If not ``None``, the standardized partial AUC over the range [0, max_fpr] is returned.
+        :param ignore_value: value to ignore if exists in target in auc calculation
         :return auc Receiver operating characteristic score
         """
         y_score = np.asarray(pred)
         y_true = np.asarray(target)
+        y_true = y_true[y_true != ignore_value]
+        y_score = y_score[y_true != ignore_value]
```
mosheraboh, line 86: Please also:
if sample_weight is not None: convert it to numpy array and then filter to be aligned with y_true

mosheraboh, line 86: Do y_true != ignore_value once

mosheraboh (changes requested):
Looks good, comments inline

fuse/eval/metrics/libs/classification.py
itaijj: I discard it

itaijj: I discard it

PR #298, opened 2023-03-23
cindex: make pred np.array to allow negation

michalozeryflato:


fuse/eval/metrics/libs/survival.py
```diff
@@ -22,6 +22,8 @@ def c_index(
         """
         if isinstance(pred[0], np.ndarray):
             pred = np.asarray(pred)[:, event_class_index]
+        elif isinstance(pred, list):
+            pred = np.asarray(pred)
```
mosheraboh, line 26: pred is a list of what? floats? 
Should we add an assert that event_class_index is -1?

michalozeryflato: pred should contain float score per sample.
The ground truth has two components: event/censor indicator and time of event/censor

michalozeryflato: @mosheraboh see new fix - following your comments

mosheraboh (approved):
Looks good

mosheraboh:
@michalozeryflato , the PR is waiting for you to merge.

PR #304, opened 2023-04-20
`mypy` typing fix

SagiPolaczek:
# Ready for CR 🕺🏼

Fixing **ALL** fuse's `mypy` errors.

### Goal:
`mypy .` runs OK on Fuse without excluding any file.

#### Note that I'm doing it as a 'side task'


mosheraboh (approved):
Great! Thanks!

PR #311, opened 2023-05-09
add a flag to metric to avoid multigpu collection

mosheraboh:


SagiPolaczek (approved):
LGTM, Thanks! 

Three general comments that come to my mind:
1. Will be interesting to measure the speedup.
2. In the future it might be cool to support partial multi gpu metric collection. We'll be able to control the tradeoff between the running time and the metric precision.
3. I guess we should start considering multi-gpu unit testing soon 😃

mosheraboh:
> LGTM, Thanks!
> 
> Three general comments that come to my mind:
> 
> 1. Will be interesting to measure the speedup.
> 2. In the future it might be cool to support partial multi gpu metric collection. We'll be able to control the tradeoff between the running time and the metric precision.
> 3. I guess we should start considering multi-gpu unit testing soon 😃

Thanks!
my main usage was to avoid collecting tricky cases.

PR #312, opened 2023-05-09
Fix ISIC's ResNet run (Poriya)

SagiPolaczek:
# Ready ✅

* Small changes in the ISIC runner
* Split ISIC test case to run on each model (Transformer, Inception, ResNet18) for better coverage
* Now `dataset_balanced_division_to_folds(reset_split=False)` has an option to verify if the total number of samples in the folds that were loaded from the `split_file` is equal to the number of samples in the given dataset.
This is useful if the user toggle the number of samples while still saving the `split_file`.

### Note
Each of the new tests takes a little more than 1 minute. So for now I think it worth it. 
Screenshot 2023-05-10 at 13 05 28


mosheraboh (dismissed):
LGTM!
Just add meaningful error message in split

PR #313, opened 2023-05-15
Prep for long train

mosheraboh:


fuse/dl/lightning/pl_funcs.py
```diff
@@ -137,7 +136,7 @@ def model_default_callbacks(
             callbacks.append(model_checkpoint_display)
 
     # last epoch checkpoint
-    callbacks.append(ModelCheckpoint(dirpath=model_dir, filename="last_epoch", save_last=True))
+    callbacks.append(ModelCheckpoint(dirpath=model_dir, save_last=True, save_top_k=0))
```
YoelShoshan, line 139: save_top_k=1 could be a better default

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
```diff
@@ -66,6 +66,10 @@ def _perplexity_update(
 
     assert len(preds.shape) == 3, f"Error: expected num dims is 3, got shape {preds.shape}"
     assert len(target.shape) == 2, f"Error: expected num dims is 2, got shape {target.shape}"
+    # to save GPU memory
+    preds = preds.detach()
```
YoelShoshan, line 70: Do you prefer to keep them on GPU?
would moving them to CPU make sense? (to free GPU mem)

detach() good but probably only helpful to avoid gradients calculation (and activations storage)

YoelShoshan (commented):
see inline

fuse/dl/lightning/pl_funcs.py
mosheraboh: Why?
(We use several callbacks, one to save the last checkpoint and others to monitor metrics/losses)

fuse/dl/lightning/pl_module.py
```diff
@@ -191,7 +193,8 @@ def predict_step(self, batch_dict: NDict, batch_idx: int) -> dict:
     def on_train_epoch_end(self) -> None:
         step_outputs = self._training_step_outputs
         # for the logs to be at each epoch, not each step
-        self.log("step", float(self.current_epoch), on_epoch=True, sync_dist=True)
+        if self._log_unit == "epoch":
```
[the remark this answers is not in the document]
mosheraboh: I will add verification, thanks,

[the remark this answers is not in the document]
mosheraboh: Done

PR #320, opened 2023-06-11
Poriya Checkpointing Aid

SagiPolaczek:
# 👁️ 
##### There are minor changes that might be merged later

@simona-rc 

Added changes for the ISIC's runner which support: **(updated 5.8.23)**
1. Loading model's backbone (ResNet18 & Transformer) prior training stage - if exists.
2. Saving model's backbone after training stage.
3. Loading model's backbone for inference stage.

### NOTE
1. I played a little with it and **didn't find it too useful for the "golden" subset case**. I guess that's because I didn't load the header weights - which should also be the case for Poriya.
2. Also **works OK for different image resolutions** since the ResNet backbone extracts a feature vector of size 512 regardless the input image dimensions (see `layer4`)
4. **Doesn't suppose to pass CICD**

## Update (5.8.23)
Now also supports checkpointing for the Transformer backbone type.


SagiPolaczek:
Closing it.
Still can be referenced if needed :) 

FYI @simona-rc

SagiPolaczek:
@simona-rc 

Reopen:
Now ISIC supports also transformer's checkpointing.

PR #324, opened 2023-06-29
add an option in CollateDefault to add fixed items to batch_dict

mosheraboh:


fuse/utils/ndict.py
```diff
@@ -144,7 +144,7 @@ def values(self) -> dict_items:
     def items(self) -> dict_items:
         return self._stored.items()
 
-    def merge(self, other: NDict) -> None:
+    def merge(self, other: MutableMapping) -> None:
```
SagiPolaczek, line 147: Now that I see it:

Why do we even need `merge()` if we just use the `update()` (which compatible regular dict..)

fuse/dl/lightning/pl_module.py
```diff
@@ -195,13 +195,15 @@ def validation_step(
         else:
             losses = self._losses
         # given the batch_dict and FuseMedML style losses - compute the losses, return the total loss (ignored) and save losses values in batch_dict["losses"]
-        _ = step_losses(losses, batch_dict)
+        if losses is not None:
```
SagiPolaczek, line 198: How `losses` can be `None`? And why the same change shouldn't be in `train` as well?

I thought that we must define one of the both: `self._validation_losses`, `self._losses`. 

Also we have the line:
```python 
        self._losses = losses if losses is not None else {}
```

So maybe we can do the same for `_validation_losses` (or assert that one of them exists?)

SagiPolaczek: Just throwing thoughts.. I'm not familiar with the motivation :)

SagiPolaczek (dismissed):
Thanks!

I added couple of questions and thoughts inline. 
besides that, LGTM 🚀

fuse/utils/ndict.py
SagiPolaczek: In the same context, I've send you a cool library on Slack that might be a cool replacement for the loved `NDict`

mosheraboh: 🤷 
Lets keep it for backward compatibility though
Saw it,
We started with this library btw, but then we decided to move to NDict cause it was with too many capabilities for our needs.

fuse/dl/lightning/pl_module.py
mosheraboh: loss is a must to train a model (and not for validation).
The code is confusing right now, let me make it clearer. Thanks for the feedback!

fuse/utils/ndict.py
SagiPolaczek: Cool! good to know! 

Indeed the `NDict` is pretty slim and simple (and thus fast).

PR #326, opened 2023-07-03
OpFunc: input=None passes the entire sample_dict

michalozeryflato:
add the option for input=None. In this case a field named "sample_dict" will be added, pointing to the entire sample_dict


mosheraboh (approved):
LGTM
FYI, if you have a func that gets sample_dict as an input, you can put it directly in the data pipeline without wrapping it.

PR #328, opened 2023-07-09
Support a better string representation for `NDict`

SagiPolaczek:
Now we can get the tree-like representation of the `NDict` also a string.

Useful for logging the cfg structure using `logging` package. For example:

```python
@hydra.main(version_base="1.2")
def main(cfg: DictConfig) -> None:
    cfg = hydra.utils.instantiate(cfg)
    log = logging.getLogger("Fuse")
    log.info("\n"+NDict(cfg).get_tree(True))
```


SagiPolaczek:
@CodiumAI-Agent

fuse/utils/tests/test_ndict.py
```diff
@@ -246,7 +246,15 @@ def test_print_tree(self) -> None:
         ndict["a.d"] = 23
         ndict["e"] = 10
 
-        ndict.print_tree(print_values=True)
+        gts = [
```
SagiPolaczek, line 249: All four possible permutations of the result (I think it isn't necessarily ordered)

mosheraboh (approved):
LGTM

PR #329, opened 2023-07-16
add ignore_index to auc_roc metric

michalozeryflato:


fuse/eval/metrics/libs/classification.py
```diff
@@ -36,6 +36,7 @@ def auc_roc(
         target: Sequence[Union[np.ndarray, int]],
         sample_weight: Optional[Sequence[Union[np.ndarray, float]]] = None,
         pos_class_index: int = -1,
+        ignore_index: int = None,
```
mosheraboh, line 39: ignore_index: Optional[int] = None,
Can you please add it in the comments as well?

mosheraboh (dismissed):
Looks good! Indeed it was missing.
Single comment inline.

fuse/eval/metrics/libs/classification.py
michalozeryflato: done

mosheraboh (approved):
Looks great!
Sorry for the long delay 😅

PR #330, opened 2023-08-01
clearml logger + improved mnist starter

shatz01:
`start_clearml_logger` now uses RANK
`simple_mnist_example` now uses jsonargparse which is prettier and really flexible
`requirements.txt` added jsonargparse


mosheraboh (approved):
Looks great, thanks!

mosheraboh (approved):
LGTM

PR #332, opened 2023-08-14
add a method to add sample ids to a dataset

michalozeryflato:
supporting extending a dataset with new samples


fuse/data/datasets/dataset_default.py
```diff
@@ -193,6 +193,13 @@ def get_all_sample_ids(self) -> List[Any]:
     def set_final_sample_ids(self, final_sample_ids: List[Any]) -> None:
         self._final_sample_ids = final_sample_ids
 
+    def add_sample_ids_and_recreate(
+        self, added_sample_ids: List[Any], **kwargs: Dict
+    ) -> None:
+        self._orig_sample_ids = self._final_sample_ids + added_sample_ids
```
SagiPolaczek, line 199: I suggest to use `self.get_all_sample_ids()` instead of `self._final_sample_ids` :)

SagiPolaczek (approved):
LGTM, Thanks!

Add a minor suggestion inline 😄

fuse/data/datasets/dataset_default.py
michalozeryflato: I think it is clearer to use _final_sample_ids, and it may be easier to track its use if the logic changes. Anyway, if you insist I will make your suggest change

PR #334, opened 2023-08-28
Bfloat16 support

mosheraboh:


fuse/eval/metrics/metrics_common.py
```diff
@@ -135,7 +135,10 @@ def collect(self, batch: Dict) -> None:
                 for name, key in self._keys_to_collect.items():
                     value = sample[key]
                     if isinstance(value, torch.Tensor):
-                        value = value.detach().cpu().numpy()
+                        value = value.detach()
```
SagiPolaczek, line 138: To optimize GPU usage we might wanna move the tensor to the CPU before changing the types:

```python
value = value.detach().cpu()
    if value.dtype == torch.bfloat16:
        value = value.to(torch.float)
value = value.numpy()
```

Just a thought.. It's pretty much negligible

fuse/dl/lightning/pl_funcs.py
```diff
@@ -240,10 +251,15 @@ def epoch_end_compute_and_log_losses(
...
                 losses[key] = []
-            if isinstance(elem[key], torch.Tensor):
-                losses(key).extend(elem[key].detach().cpu().tolist())
+            value = elem[key]
+            if isinstance(value, torch.Tensor):
+                value = value.detach()
+                if value.dtype == torch.bfloat16:
+                    value = value.to(torch.float)
+                value = value.cpu()
+                losses(key).extend(value.tolist())
```
SagiPolaczek, line 260: Weird, shouldn't it be with `[ ]`? 
i.e:
```python
losses[key].extend(value.tolist())
```
Otherwise we should get an error:
```
TypeError: 'dict' object is not callable
```

WDYT?

SagiPolaczek (dismissed):
two comments inline :)

PR #340, opened 2024-01-18
Add OpReplaceAnyElements to ops_common

sivanravidos:
- Added OpReplaceAnyElements to opps_common. This op allow replacing any set of tokens with a given token.
This is useful for example when one wants to ingore a set of tokens, e.g. when applying a metric that should ignore all special tokens.
- Added unit test for the new op and the former OpReplaceElements

### DIFF
--- a/fuse/data/ops/ops_common.py
+++ b/fuse/data/ops/ops_common.py
@@ -718,7 +718,7 @@ class OpReplaceAnyElements(OpBase):
         key_in: str,
         find_any_val: List[Any],
         replace_with_val: Any,
-        key_out: str = None,
+        key_out: Optional[str] = None,
     ) -> NDict:
         """
         Args:
--- a/fuse/data/ops/ops_common.py
+++ b/fuse/data/ops/ops_common.py
@@ -702,3 +702,56 @@ class OpReplaceElements(OpBase):
             raise Exception(f"Unsupported object type {type(input_obj)}")
 
         return sample_dict
+
+class OpReplaceAnyElements(OpBase):
+    """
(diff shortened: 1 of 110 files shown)


fuse/data/ops/ops_common.py
```diff
@@ -702,3 +702,61 @@ def __call__(
             raise Exception(f"Unsupported object type {type(input_obj)}")
 
         return sample_dict
+
+
+class OpReplaceAnyElements(OpBase):
+    """
+    Replace any element from a given list of elements with a requested value
```
mosheraboh, line 709: list./ numpy array or tensor

fuse/data/ops/ops_common.py
```diff
@@ -702,3 +702,61 @@ def __call__(
...
+    This is useful, for example, when converting a set of token IDs to -100 to make a metric ignore certain elements (e.g. all special tokens).
+    """
+
+    def __call__(
+        self,
+        sample_dict: NDict,
+        key_in: str,
+        find_any_val: List[Any],
+        replace_with_val: Any,
+        key_out: str = None,
```
mosheraboh, line 721: Optional[str]

mosheraboh (approved):
Looks good.
Minors inline.

fuse/data/ops/ops_common.py
sivanravidos: fixed

fuse/data/ops/ops_common.py
sivanravidos: fixed

PR #342, opened 2024-01-18
Make `NDict` (hopefully) lighter to import

SagiPolaczek:
NOTE:

It might be the case that `from torch import Tensor` still runs "unnecessary" code.


mosheraboh (approved):
LGTM
Lets check if we can avoid this import how.

PR #343, opened 2024-02-08
Crop padding in collate

liamhazan:


fuse/data/utils/collates.py
```diff
@@ -190,3 +190,41 @@ def _pad_size(value: torch.Tensor, dim: int) -> List[int]:
             padded_values.append(padded_value)
 
         return default_collate(padded_values)
+
+    @staticmethod
+    def crop_padding(
+        input_ids_list: List[torch.Tensor], pad_token_id: int
+    ) -> torch.Tensor:
+        """
+        Crop padding of a batch of input_ids tensors to the minimum length possible.
```
mosheraboh, line 199: Can you specify here what is the expected/ supported dim of an element in input_ids? I assume 1d tensors only.

mosheraboh (commented):
Looks good!
One comment inline.

mosheraboh (approved):

PR #347, opened 2024-03-27
shared memory utility helping to speedup reading from files significantly if enough RAM is available 

YoelShoshan:
shared memory utility helping to speedup reading from files significantly if enough RAM is available 

seen at least x10 speedup.

on a machine that is already bottlenecked on disk seen over x100 speedup


fuse/data/utils/collates.py
```diff
@@ -134,7 +134,8 @@ def _batch_dispatch(
             # use special handler if specified
             batch_dict[key] = self._special_handlers_keys[key](collected_values)
         elif isinstance(
-            collected_values[0], (torch.Tensor, np.ndarray, float, int, str, bytes)
+            collected_values[0],
```
mosheraboh, line 137: Do you really need it?
It will collect each element separately 
If you do it - I suggest adding a flag, cause sometimes we want to avoid it.
Or maybe add a special handler whenever you want this behavior.

fuse/utils/multiprocessing/linux_shared_memory.py
```diff
@@ -0,0 +1,92 @@
...
+# from multiprocessing import Lock
+from filelock import FileLock
+from glob import glob
+import psutil
+
+SHARED_MEM_FILES_PREFIX = "OUR_SHARED_MEM_@"
+SHARED_MEM_ACTIVE_ENV_VAR = "ACTIVATE_OUR_SHARED_MEM"
+SHM_BASE_DIR = "/dev/shm/"
+
+G_lock = FileLock(join(SHM_BASE_DIR, "our_shared_mem_file_lock"))  # Lock()
```
mosheraboh, line 14: Why do we need a global lock? shouldn't per copied file lock work?

fuse/utils/multiprocessing/linux_shared_memory.py
```diff
@@ -0,0 +1,92 @@
...
+    if os.environ[SHARED_MEM_ACTIVE_ENV_VAR].lower() not in ["1", "t", "true"]:
+        return file_path
+
+    with G_lock:
+        assert os.path.isfile(
+            file_path
+        ), f"file_path does not point to a file: file_path={file_path}"
+
+        src_file_size_bytes = os.stat(file_path).st_size
+        dest = join(SHM_BASE_DIR, SHARED_MEM_FILES_PREFIX + basename(file_path))
```
mosheraboh, line 40: I wonder if we want to keep the entire path structure (to avoid conflicts and keep the same relative path). 
For example:
file /mnt/bmfmbio/ab_design/1.fasta
Will be copied to /dev/shm/$SHARED_MEM_FILES_PREFIX/mnt/bmfmbio/ab_design/1.fasta

mosheraboh (approved):
Great solution!
Few comments inline.

fuse/data/utils/collates.py
YoelShoshan: I'll remove it from this PR

fuse/utils/multiprocessing/linux_shared_memory.py
YoelShoshan: I can do it per file, but where should it be initialized ?
also - if we don't keep it sequential like now, the free RAM check won't be correct anymore

fuse/utils/multiprocessing/linux_shared_memory.py
YoelShoshan: was considering that - I did remember checking that /dev/shm does support directories in it

was a bit reluctant to do it because I run "remove" commands which delete files, so didn't want '..' paths to somehow escape it

YoelShoshan: trying it now

YoelShoshan: done - now full path will be stored under the shared memory directory.

PR #352, opened 2024-05-07
MetricCountSeqAndTokens counst tokens in label (if exist) + improving epoch metrics print

michalozeryflato:
1. add optional decoder_input to MetricCountSeqAndTokens
2. print traceback before throw - helps in debugging
3. modify epoch stats printing - each column has its own (fixed) width

### DIFF
--- a/fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
+++ b/fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
@@ -40,7 +40,7 @@ class MetricCountSeqAndTokens(MetricPerBatchDefault):
     ) -> None:
         """
         :param encoder_input: key to the encoder_input
-        :param decoder_input: key to the encoder_input
+        :param decoder_input: key to the decoder_input
         :param ignore_index: token_id to ignore (not to count), typically pad token id
         :param state: the sequence count and token count to continue for. Should be restored when we continue training.
                     use get_state() to get the state and save it upon checkpointing,
--- a/fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
+++ b/fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
@@ -203,7 +203,7 @@ def _perplexity_update(
         preds = preds[mask]
         count = mask.sum()
     else:
-        count = target.shape[0]
+        count = torch.tensor(target.shape[0])
 
     preds = torch.gather(preds, 1, target.view(-1, 1)).squeeze(1)
     # avoid from overflow
@@ -211,7 +211,6 @@ def _perplexity_update(
         preds = preds.to(torch.float32)
         preds = torch.clamp(preds, min=1e-10)
     total_log_probs = -preds.log().sum()
-    count = mask.sum()
 
     return {"log_probs": total_log_probs.unsqueeze(0), "token_num": count.unsqueeze(0)}
 
--- a/fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
+++ b/fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
@@ -153,6 +153,7 @@ class MetricPerplexity(MetricPerBatchDefault):
 
 # Copied internal function https://github.com/Lightning-AI/metrics/blob/825d17f32ee0b9a2a8024c89d4a09863d7eb45c3/src/torchmetrics/functional/text/perplexity.py#L68
 # copied and not imported to not be affected by internal interface modifications.
+# modifications: (1) reshape => view (2) apply mask at the beginning of computation (3) use torch.gather
 def _perplexity_update(
     batch_dict: dict,
     preds_key: str,
@@ -193,18 +194,18 @@ def _perplexity_update(
     preds = preds.detach()
     target = target.detach()
 
-    preds = preds.reshape(-1, preds.shape[-1])
-    target = target.reshape(-1)
+    preds = preds.view(-1, preds.shape[-1])
+    target = target.view(-1)
 
     if ignore_index is not None:
         mask = target.ne(ignore_index)
-        target = target.where(
-            target != ignore_index, torch.tensor(0, device=target.device)
-        )
+        target = target[mask]
+        preds = preds[mask]
+        count = mask.sum()
     else:
-        mask = torch.ones_like(target, dtype=torch.bool)
+        count = target.shape[0]
 
-    preds = preds[:, target].diagonal()[mask]
+    preds = torch.gather(preds, 1, target.view(-1, 1)).squeeze(1)
     # avoid from overflow
     if preds.dtype == torch.float16:
         preds = preds.to(torch.float32)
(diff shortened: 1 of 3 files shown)


fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
```diff
@@ -82,12 +85,15 @@ def get_state(self) -> dict:
...
     encoder_input_key: str,
+    decoder_input_key: Optional[str] = None,
     ignore_index: Optional[int] = None,
 ) -> Dict[str, torch.Tensor]:
     """Count number of sequences and tokens
     Args:
         encoder_input_key:
             key to encoder_input
+        decoder_input_key:
+            key to encoder_input
```
SagiPolaczek, line 96: key to *decoder_input :)

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
```diff
@@ -98,19 +104,29 @@ def _count_seq_and_tokens_update(
...
     seq_num = torch.tensor(
         mask.shape[0], dtype=torch.int64, device=encoder_input.device
     )
     token_num = mask.sum().to(dtype=torch.int64)
 
+    if decoder_input_key is not None and decoder_input_key in batch_dict:
+        decoder_input = batch_dict[decoder_input_key].detach()
+        mask2 = get_mask(decoder_input, ignore_index)
+        assert mask2.shape[0] == mask.shape[0]
+        token_num += mask2.sum().to(dtype=torch.int64)
```
SagiPolaczek, line 118: Just making sure:
Here's the support for counting the labels tokens as well?

If so, best to double check with @mosheraboh :)

michalozeryflato: @SagiPolaczek I updated @mosheraboh, but better confirm the change with him.

fuse/utils/misc/misc.py
```diff
@@ -158,22 +158,22 @@ def time_display(seconds: int, granularity: int = 3) -> str:
...
-    max_val_width = np.vectorize(len)(
-        df.values.astype(str)
-    ).max()  # get maximum length of all values
-    max_col_width = max(
-        [len(x) for x in df.columns]
-    )  # get the maximum lengths of all column names
-    col_width = max(max_col_width, max_val_width, col_width)
-
-    dashes = (col_width + 2) * len(df.columns.values)
+    max_val_width_per_col = np.vectorize(len)(df.values.astype(str)).max(
```
mosheraboh, line 161: Here you've modified the code to set the width per column instead of a single width for all columns?

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
mosheraboh: ☝️

mosheraboh (approved):
LGTM

fuse/utils/misc/misc.py
michalozeryflato: yes, to make the overall width smaller - so it better fits the screen

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
```diff
@@ -177,18 +194,18 @@ def _perplexity_update(
     preds = preds.detach()
     target = target.detach()
 
-    preds = preds.reshape(-1, preds.shape[-1])
-    target = target.reshape(-1)
+    preds = preds.view(-1, preds.shape[-1])
```
mosheraboh, line 197: why? to save memory?

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
```diff
@@ -137,6 +153,7 @@ def __init__(
 
 # Copied internal function https://github.com/Lightning-AI/metrics/blob/825d17f32ee0b9a2a8024c89d4a09863d7eb45c3/src/torchmetrics/functional/text/perplexity.py#L68
 # copied and not imported to not be affected by internal interface modifications.
+# modifications: (1) reshape => view (2) apply mask at the beginning of computation (3) use torch.gather
```
mosheraboh, line 156: Nice solution.

mosheraboh (approved):
Looks great

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
michalozeryflato: @mosheraboh From the documentation of [reshape](https://pytorch.org/docs/stable/generated/torch.reshape.html#torch.reshape): When possible, the returned tensor will be a **view of input. Otherwise, it will be a copy**. Contiguous inputs and inputs with compatible strides can be reshaped without copying, but you should not depend on the copying vs. viewing behavior.

Question is: 
(1) do we assume view will always work, and otherwise alert the user. Or 
(2) do we want to be permissive and copy the data (despite additional GPU memory) when view is not possible (i.e. memory is not contiguous)

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
```diff
@@ -177,24 +194,23 @@ def _perplexity_update(
...
-    target = target.reshape(-1)
+    preds = preds.view(-1, preds.shape[-1])
+    target = target.view(-1)
 
     if ignore_index is not None:
         mask = target.ne(ignore_index)
-        target = target.where(
-            target != ignore_index, torch.tensor(0, device=target.device)
-        )
+        target = target[mask]
```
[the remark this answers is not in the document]
SagiPolaczek: I'm not sure it's critical (maybe it's even better, saving memory)

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
SagiPolaczek: Is it really that expensive in memory?

If it is, than I would warn the user conditionally to `Tensor.is_contiguous()` result :)

SagiPolaczek (dismissed):
THANK YOU!

Two comments inline

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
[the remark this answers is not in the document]
michalozeryflato: yes, I have reduced the side of pred and target, since at the end the entries corresponding to the mask are dropped. So it seems to produce the same results, but with less memory/time (not having to process the masked entries)

fuse/eval/metrics/sequence_gen/metrics_seq_gen_common.py
```diff
@@ -177,24 +194,25 @@ def _perplexity_update(
     preds = preds.detach()
     target = target.detach()
 
+    # reshape attempts to create a view, and COPIES the data if fails.
+    # an issue to consider:  use view and alert the user if it fails?
     preds = preds.reshape(-1, preds.shape[-1])
     target = target.reshape(-1)
 
     if ignore_index is not None:
         mask = target.ne(ignore_index)
```
SagiPolaczek, line 203: I guess we can use `get_mask()` here

SagiPolaczek (approved):
LGTM! Thanks!

PR #354, opened 2024-05-22
min_size_per_dim  + changes for scalars inputs+outputs support

YoelShoshan:
by allowing min_size_per_dim to be longer than the number of dims in a tensor, it makes usage easier.

for example, 

min_size_per_dim=(10_000, -1)
would then work for tensors of single dim - e.g. of size (1234, )
and also for tensors of 2(+) dims - e.g. of size (4321, 100)

this is much more nice for the user in several cases.

also - few changes for scalars inputs+outputs


mosheraboh (approved):
LGTM

fuse/data/ops/ops_cast.py
```diff
@@ -37,12 +37,15 @@ def to_tensor(
         value: Any,
         dtype: Optional[torch.dtype] = None,
         device: Optional[torch.device] = None,
+        ignore_None: bool = False,
```
mosheraboh, line 40: ignore_none

fuse/eval/metrics/libs/stat.py
```diff
@@ -20,12 +20,22 @@ def pearson_correlation(
         :param target: target values
         :param mask: optional boolean mask. if it is provided, the metric will be applied only to the masked samples
         """
+        assert len(pred) > 0
         if isinstance(pred, Sequence):
-            pred = np.array(pred)
+            if np.isscalar(pred[0]):
+                pred = np.array(pred)
+            else:
+                pred = np.concatenate(pred)
```
mosheraboh, line 28: here pred is numpy? necessarily same dims?

mosheraboh (dismissed):
Looks good.
Minor comments inline.

fuse/eval/metrics/libs/stat.py
YoelShoshan: yes, in the else it's numpy, and it not necessarily same dims (it's 1d each)

PR #356, opened 2024-05-29
implement hdf5 dataset reader op

mosheraboh:


fuse/data/ops/ops_read.py
```diff
@@ -113,3 +114,51 @@ def get_all_keys(self) -> List[Hashable]:
         :return: list of  dataframe index values
         """
         return list(self.data.keys())
+
+
+class OpReadHDF5(OpBase):
+    """
+    Op reading data from hd5f based dataset
```
SagiPolaczek, line 121: `hd5f` typo :)

fuse/data/ops/ops_read.py
```diff
@@ -113,3 +114,51 @@ def get_all_keys(self) -> List[Hashable]:
...
+        :param columns_to_extract: list of columns to extract - dataset keys to extract. When None (default) all columns are extracted
+        :param rename_columns: rename columns
+        :param key_index: name of value in sample_dict which will be used as the key/index
+        :param key_column: name of the column which use as key/index. In case of None, the original dataframe index will be used to extract the values for a single sample.
+        """
+        super().__init__()
+        # store input
+        self._data_filename = data_filename
+        self._columns_to_extract = columns_to_extract
+        self._rename_columns = rename_columns if rename_columns is not None else {}
```
SagiPolaczek, line 142: Nice :)

SagiPolaczek: (avoiding mutable default values)

fuse/data/ops/ops_read.py
```diff
@@ -113,3 +114,51 @@ def get_all_keys(self) -> List[Hashable]:
...
+        self._key_index = key_index
+        self._key_column = key_column
+
+        self._h5 = h5py.File(self._data_filename, "r")
```
SagiPolaczek, line 146: Seeing that I was concern that we load the entire file into memory, but seems not:
https://stackoverflow.com/questions/40449659/does-h5py-read-the-whole-file-into-memory

fuse/data/ops/ops_read.py
```diff
@@ -113,3 +114,51 @@ def get_all_keys(self) -> List[Hashable]:
...
+
+        if self._columns_to_extract is None:
+            self._columns_to_extract = list(self._h5.keys())
+
+        self._num_samples = len(self._h5[self._columns_to_extract[0]])
```
SagiPolaczek: Hope this line doesn't load it into memory

fuse/data/ops/ops_read.py
SagiPolaczek: I think we can use frozen dict here as well

fuse/data/ops/ops_read.py
```diff
@@ -113,3 +114,51 @@ def get_all_keys(self) -> List[Hashable]:
...
+
+    def num_samples(self) -> int:
+        return self._num_samples
```
SagiPolaczek: I don't see where you use it

SagiPolaczek (approved):
Not sure if relevant, but just mentioning we have `fuse.utils.file_io.file_io.load_hdf5()` and `fuse.utils.file_io.file_io.save_hdf5_safe()` - maybe you'll find it relevant (tho I see the code here is cleaner)

PR #358, opened 2024-06-25
Matching keys lengths and new metrics

IdoAmosIBM:
Added 2 new metrics: mathews correlation coef. and balanced accuracy.
Added a new functionality to default collate allowing to match the length of some keys in the batch dict to other keys - example: for encoder model, matching the length of the labels to the length of the input after crop padding is applied.

### DIFF
--- a/fuse/data/utils/collates.py
+++ b/fuse/data/utils/collates.py
@@ -66,12 +66,7 @@ class CollateDefault(CollateToBatchList):
         self._keep_keys = keep_keys
         self._add_to_batch_dict = add_to_batch_dict
 
-        self._post_collate_special_handlers_keys = None
-        if post_collate_special_handlers_keys is not None:
-            self._post_collate_special_handlers_keys = {}
-            self._post_collate_special_handlers_keys.update(
-                post_collate_special_handlers_keys
-            )
+        self._post_collate_special_handlers_keys = post_collate_special_handlers_keys
 
     def __call__(self, samples: List[Dict]) -> Dict:
         """
@@ -117,8 +112,8 @@ class CollateDefault(CollateToBatchList):
             batch_dict.update(self._add_to_batch_dict)
 
         if self._post_collate_special_handlers_keys is not None:
-            for key in self._post_collate_special_handlers_keys:
-                self._post_collate_special_handlers_keys[keys](batch_dict)
+            for callable in self._post_collate_special_handlers_keys:
+                callable(batch_dict)
 
         return batch_dict
 
--- a/fuse/data/utils/collates.py
+++ b/fuse/data/utils/collates.py
@@ -257,7 +257,7 @@ class CollateDefault(CollateToBatchList):
         return batched_sequences
 
     @staticmethod
-    def match_length_to_target_key(
+    def crop_length_to_match_target_key(
         batch_dict: dict, target_key: str, keys_to_match: List[str]
     ) -> None:
         """
--- a/fuse/data/utils/collates.py
+++ b/fuse/data/utils/collates.py
@@ -43,6 +43,7 @@ class CollateDefault(CollateToBatchList):
         keep_keys: Sequence[str] = tuple(),
         raise_error_key_missing: bool = True,
         special_handlers_keys: Optional[Dict[str, Callable]] = None,
+        post_collate_special_handlers_keys: Optional[List[Callable]] = None,
         add_to_batch_dict: Optional[Dict[str, Any]] = None,
     ):
         """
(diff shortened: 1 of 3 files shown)


fuse/eval/metrics/metrics_common.py
```diff
@@ -160,7 +160,10 @@ def collect(self, batch: Dict) -> None:
             batch_to_collect = {}
 
             for name, key in self._keys_to_collect.items():
-                value = batch[key]
+                try:
```
mosheraboh, line 163: Was it just for debugging?
Or would you like to print a more informative message and then reraise the exception?

fuse/eval/metrics/classification/metrics_classification_common.py
```diff
@@ -336,3 +337,81 @@ def __init__(self, pred: str, target: str, **kwargs: dict):
...
+            pred=pred,
+            target=target,
+            sample_weight=sample_weight,
+            metric_func=self.mcc_wrapper,
+            **kwargs,
+        )
+
+    def mcc_wrapper(
+        self,
+        pred: Optional[str] = None,
```
mosheraboh, line 368: is it a string or a list of numpy array?
Are you expecting class predictions or scores after sofrmax?

fuse/data/utils/collates.py
```diff
@@ -247,3 +255,22 @@ def crop_padding(
         cropped_sequences = [ids[:min_length] for ids in input_ids_list]
         batched_sequences = torch.stack(cropped_sequences, dim=0)
         return batched_sequences
+
+    @staticmethod
+    def match_length_to_target_key(
```
mosheraboh, line 260: Cropping?
What happens when the target_key is shorter then keys_to_match?

fuse/data/utils/collates.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         keep_keys: Sequence[str] = tuple(),
         raise_error_key_missing: bool = True,
         special_handlers_keys: Optional[Dict[str, Callable]] = None,
+        post_collate_special_handlers_keys: Optional[List[Callable]] = None,
```
mosheraboh, line 46: Why not to use special_handlers_keys instead?
Can you give an example of a use case?

mosheraboh (commented):
Thanks Ido!
Few comments and questions inline.

fuse/data/utils/collates.py
IdoAmosIBM: this will raise an error, maybe I'll change the name length to `crop_length_to_target_key`? The motivation is when the encoder and labels should have the same length and pad cropping was applied to the encoder input - so we want to match the labels.

fuse/data/utils/collates.py
IdoAmosIBM: a use case is `match_target_key_to_length` from below, we can't use special handlers because it handles one element in the batch dict each time, we need to match one key to another after the special handler has been applied.

fuse/data/utils/collates.py
mosheraboh: I assume you mean decoder input?
I asked Michal to solve it in a different way.
(labels and decoder should always have the same length).

fuse/data/utils/collates.py
mosheraboh: Why is it a list and not a single callable?

fuse/data/utils/collates.py
IdoAmosIBM: sorry, I do mean encoder, this is for the case of an encoder-only model

fuse/eval/metrics/classification/metrics_classification_common.py
```diff
@@ -336,3 +337,81 @@ def __init__(self, pred: str, target: str, **kwargs: dict):
...
+            pred=pred,
+            target=target,
+            sample_weight=sample_weight,
+            metric_func=self.balanced_acc_wrapper,
+            **kwargs,
+        )
+
+    def balanced_acc_wrapper(
+        self,
+        pred: Union[List, np.ndarray],
```
mosheraboh, line 407: It's still not clear if you expect here sofrmax scores (we call pred) or class predictions (we call cls_pred).

mosheraboh (dismissed):
Thanks Ido!
I still have few comments.
In order not to block you, I'm ok with merging and potentially make some modifications in a new PR.

fuse/data/utils/collates.py
IdoAmosIBM: to allow multiple use cases, e.g. aligning a set of keys to key A and a different set of keys to key B.

fuse/eval/metrics/classification/metrics_classification_common.py
IdoAmosIBM: oh I understand, I expect class predictions I used the same names as in MetricAccuracy. Do you want me to change to cls_pred?

SagiPolaczek (approved):
LGTM! 
Thanks!

PR #362, opened 2024-08-11
Support tuples as sample IDs  in Bootstrapping

ellabarkan:


fuse/eval/metrics/metrics_common.py
```diff
@@ -887,6 +887,23 @@ def reset(self) -> None:
         self._metric.reset()
         return super().reset()
 
+    @staticmethod
+    def _convert_tuples(ids: List[Tuple[str, int]]) -> np.ndarray:
```
mosheraboh, line 891: Are you assuming tuple of str and int? 
Can we do something more general?

ellabarkan: Thx. The method of convert is supporting all tuple of everything of any length, I need to change the interface definition

fuse/eval/metrics/metrics_common.py
```diff
@@ -887,6 +887,23 @@ def reset(self) -> None:
...
+        for i, tuple_elem in enumerate(sample_tuple):
+            if isinstance(tuple_elem, str):
+                max_len = max(len(str(el[i])) for el in ids)
+                dtype_tuple.append((f"field{i}", f"U{max_len}"))
+            else:
+                dtype_tuple.append((f"field{i}", type(tuple_elem)))
+
+        ids = np.array(ids, dtype=dtype_tuple)
+        ids = [tuple(x) for x in ids]
+        return ids
```
mosheraboh, line 905: Are you returning here a list of tuples or numpy array?

fuse/eval/metrics/metrics_common.py
```diff
@@ -920,7 +941,11 @@ def eval(
                 stratum_filter = stratum_id == stratum
                 n_stratum = sum(stratum_filter)
                 random_sample = rnd.randint(0, n_stratum, size=n_stratum)
-                sampled_ids[stratum_filter] = ids[stratum_filter][random_sample]
+
+                flt_indx = np.where(stratum_filter)[0]
```
mosheraboh, line 945: You needed to move to "for loop implementation" even though you converted it ids to numpy array?

fuse/eval/metrics/metrics_common.py
ellabarkan: When applying only line 903 the results numpy array includes tuples but they have a problem of being hashable and I was getting error in the line :
permutation = [original_ids_pos[sample_id] for sample_id in required_ids] 
TypeError: unhashable type: 'writeable void-scalar'

The suggestion for the fix I found was to updated their types to tuple that is supported in the list and not in array. If I am trying to convert them after to np.array again the tuples are broken to separate elements and we having 2D array

sivanravidos:
@ellabarkan @mosheraboh 
I have the same issue I think when calling GroupAnalysis or Filter, who uses the Collector
in line 410:
```permutation = [original_ids_pos[sample_id] for sample_id in required_ids]```

`original_ids_pos` has tuples and `required_ids` has ndarrays.

I added this hack just to use the code in the meantime
``` permutation = [original_ids_pos[(sample_id[0], int(sample_id[1]))] for sample_id in required_ids]```

Can we add solving this too to this PR and also finalizing it?


<br>
Screenshot 2024-11-18 at 20 26 56

sivanravidos:
@mosheraboh I checked and on this branch I get the same exception on line 410, this solution doesn't fix the get method

sivanravidos:
@mosheraboh Ella and I discussed again, the problem is as follows:
- we use Fuse wrapper that wraps dataset ids as tuples `(dataset_name, id)`
(dataset_wrap_seq_to_dict.py line 104)
- Tuple arrays are problematic for some operations, you can't apply boolean filters on them,
this happens in several places, specificaly in CI, Filter and GroupAnalysis so we need to fix everywhere
- see an alternative generic fix I hope, using list comprehension instead of filtering in branch sample_id_fix (fixed my prev suggestion):
https://github.com/BiomedSciAI/fuse-med-ml/compare/group_analyis_configuration..sample_id_fix

PR #363, opened 2024-08-26
clearml offline mode

mosheraboh:


fuse/dl/lightning/pl_module.py
```diff
@@ -176,6 +176,9 @@ def __init__(
 
     ## forward
     def forward(self, batch_dict: NDict) -> NDict:
+        # workaround for fsdp
+        if not isinstance(batch_dict, NDict):
```
YoelShoshan, line 180: is there a point to run some assertion that it's at least a dictionary? or do you think that NDict covers that enough already anyway?

mosheraboh: It should be ok

PR #364, opened 2024-09-10
added new metrics for regression tasks

IdoAmosIBM:


fuse/eval/metrics/libs/stat.py
```diff
@@ -55,3 +55,57 @@ def pearson_correlation(
...
+            raise ValueError(
+                f"expected 1D vectors. got pred shape: {pred.shape}, target shape: {target.shape}"
+            )
+
+        assert len(pred) == len(
+            target
+        ), f"Spearman corr expected to get pred and target with same length but got pred={len(pred)} - target={len(target)}"
+
+        statistic, p_value = spearmanr(
+            pred, target, nan_policy="propagate"
```
mosheraboh, line 105: Should we expose nan_policy to be an argument in metric class?

fuse/eval/metrics/stat/metrics_stat_common.py
```diff
@@ -31,5 +34,118 @@ def __init__(
             target=target,
             mask=mask,
             metric_func=Stat.pearson_correlation,
-            **kwargs
+            **kwargs,
+        )
+
+
+class MetricSpearmanCorrelation(MetricDefault):
```
mosheraboh, line 41: I suggest that all of those metrics (including  MetricPearsonCorrelation) be put in a new folder for regression. Makes sense?

mosheraboh (commented):
Looks great!
Minors inline.

fuse/eval/metrics/libs/stat.py
IdoAmosIBM: it made sense to me to make this visibility to understand where nans might come from but i don't mind removing

fuse/eval/metrics/stat/metrics_stat_common.py
IdoAmosIBM: yes makes a lot of sense, will refractor

YoelShoshan:
FYI

@IdoAmosIBM @mosheraboh 
Note that as a result of this PR some tests in dependent libs failed, as it relied on imports from a certain sub dir/module.
For backward compatibility I'm making them accessible from the same location (do tell me if you prefer something else)

here's the PR - https://github.com/BiomedSciAI/fuse-med-ml/pull/365

IdoAmosIBM:
I don?t see an option to approve the PR so will have to be @Moshiko ***@***.***>

From: YoelShoshan ***@***.***>
Date: Wednesday, 11 September 2024 at 15:09
To: BiomedSciAI/fuse-med-ml ***@***.***>
Cc: Ido Amos ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [BiomedSciAI/fuse-med-ml] added new metrics for regression tasks (PR #364)
FYI @IdoAmosIBM @mosheraboh Note that as a result of this PR some tests in dependent libs failed, as it relied on imports from a certain sub dir/module. For backward compatibility I'm making them accessible from the same location (do tell me


FYI

@IdoAmosIBM<https://github.com/IdoAmosIBM> @mosheraboh<https://github.com/mosheraboh>
Note that as a result of this PR some tests in dependent libs failed, as it relied on imports from a certain sub dir/module.
For backward compatibility I'm making them accessible from the same location (do tell me if you prefer something else)

here's the PR - #365<https://github.com/BiomedSciAI/fuse-med-ml/pull/365>

?

PR #366, opened 2024-09-12
move modular tokenizer to fuse

mosheraboh:


SagiPolaczek (approved):
LGTM.

Note that we don't test it at the moment

PR #367, opened 2024-09-15
support custom configure_optimizers method and avoid from saving random model by default

mosheraboh:


SagiPolaczek (dismissed):
Thanks! LGTM :)

I would consider to change the type hint of `optimizers_and_lr_schs` into:

```
        optimizers_and_lr_schs: Union[dict,callable] = None,
```

PR #372, opened 2024-10-04
rmse score fix

mosheraboh:


SagiPolaczek (approved):
Nice catch and neat fix 😄

SagiPolaczek:
The CICD fails to import the new metrics, but it works fine on my env.. 
I guess it relates to the version:
Screenshot 2024-10-04 at 17 20 28

PR #373, opened 2024-10-06
Add new tokens to tokenizer op

matanninio:
Added a proxy function to allow adding special tokens to the tokenizer via the tokenizer_op.

Also minor fix - added "r'..'" to a regexp string


SagiPolaczek (approved):
LGTM :)

PR #374, opened 2024-10-07
advancing to scalars gen2

YoelShoshan:


fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -147,10 +144,12 @@ def prepare_info_for_model_step(
...
-        scalars_indices = []
-        scalars_values = []
-        scalars_masked_indices = []
-        prev_index_end = -1
+        ## both `all_scalars_values` and `all_scalars_valid_mask` will contain torch tensors, which will be concatanated in the end of this function
+        all_scalars_values = (
+            []
+        )  # one scalar for every element, `scalar_default_unfound_value` is used for elements that aren't scalars
+        all_scalars_valid_mask = []  # for each element, whether it's a scalar or not
+        scalar_default_unfound_value = -1000.0
```
mosheraboh, line 152: What does it mean?

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -47,7 +45,7 @@ class InjectorToModularTokenizerLib:
     @staticmethod
     def build_placeholder_meta_tokenization(
         *,
-        sequence: Union[str, list, tuple],
+        sequence: str,
```
mosheraboh, line 48: Why did you change it if you still support those types?

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -91,19 +88,18 @@ def build_placeholder_meta_tokenization(
             if tokenizer_type.startswith("SCALARS_"):
                 with_placeholders.append(
                     "<@TOKENIZER-TYPE=AA>"
```
mosheraboh, line 90: The general solution can't assume there is AA subtokenizer.
Maybe we need a default empty sub-tokenizer? Maybe SCALARS can be an empty sub-tokenizer?

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -91,19 +88,18 @@ def build_placeholder_meta_tokenization(
...
-                )  # won't use AA tokens, just an arbitrary one to be able to use a token like <SCALAR>
+                )  # AA tokenizer selection is arbitrary, we only take the special token <SCALAR> from it
 
-                if (
-                    tokenizer_type == "SCALARS_LITERALS"
-                ):  # note: masking is only supported in literals (not in "from dict")
+                if tokenizer_type == "SCALARS_LITERALS":
```
mosheraboh, line 93: Remind me, can we put mask in scalar literals?

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -147,10 +144,12 @@ def prepare_info_for_model_step(
 
 
         """
-        scalars_indices = []
-        scalars_values = []
-        scalars_masked_indices = []
-        prev_index_end = -1
+        ## both `all_scalars_values` and `all_scalars_valid_mask` will contain torch tensors, which will be concatanated in the end of this function
+        all_scalars_values = (
```
mosheraboh, line 148: all_scalar_values = [] ?

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -91,19 +88,18 @@ def build_placeholder_meta_tokenization(
...
+                if tokenizer_type == "SCALARS_LITERALS":
                     values = subseq.split(",")
-                    # seq = "<SCALAR>" * len(values)
-                    seq = "".join(
-                        [
-                            "<MASKED_SCALAR>" if x == "<MASK>" else "<SCALAR>"
-                            for x in values
-                        ]
-                    )
+                    # validate that all values can be converted to fload
```
mosheraboh, line 95: typo: float

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -209,24 +196,47 @@ def prepare_info_for_model_step(
...
+                )
+
+        all_scalars_values = torch.concat(all_scalars_values)
+        all_scalars_valid_mask = torch.concat(all_scalars_valid_mask)
+
+        assert all_scalars_values.shape == all_scalars_valid_mask.shape
+
+        # pad if needed
+        full_query_len = len(token_ids)
+        if full_query_len > all_scalars_values.shape[0]:
```
mosheraboh, line 218: Can't it happend?

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -130,6 +126,7 @@ def prepare_info_for_model_step(
         *,
```
mosheraboh, line 126: maybe build_scalars be a better name here?

mosheraboh (approved):
Looks great!
Few comments inline/

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
YoelShoshan: since we now keep the scalars values and mask at the size of the entire sequence, this is the default value for places that don't actually have a scalar value.
I chose -1000 and not something like 0 to make sure it pops up easily if there are mistakes down the road

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
YoelShoshan: interesting point.
SCALARS is currently fully programmatic and does not rely on any dictionary, so I would rather not mix it.
Probably better to have "base" that gets automatically generated and supported , as the modular tokenizer already knows how to handle special tokens

maybe "Base" or "SpecialTokensBase" or something

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
YoelShoshan: I just copied over the code, so probably you already fixed this and it got overriden.
fixed it now to contain the union.

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
YoelShoshan: I think that static code auto "fixed" this, due to the comment length
(I didn't write it like this)

YoelShoshan: I moved the comments to one line above the code, so it won't do that.

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
YoelShoshan: thanks!
fixed

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
YoelShoshan: Do you mean "can it happen" ? 
If so, then yes, almost always.

The main code logic (before it) iterates over each sub part (with specific sub tokenizer) so it does not contain the padding.

I can explain more if it isn't clear

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
YoelShoshan: np, renamed

fuse/data/tokenizers/modular_tokenizer/inject_utils.py
```diff
@@ -209,24 +197,52 @@ def prepare_info_for_model_step(
...
+                [
+                    all_scalars_valid_mask,
+                    torch.full(
+                        (pad_len,), fill_value=False, dtype=all_scalars_valid_mask.dtype
+                    ),
+                ]
             )
-        else:
-            scalars_masked_indices = None
+        elif full_query_len > all_scalars_values.shape[0]:
```
YoelShoshan, line 239: @mosheraboh see here, related to what we talked about.

I'll try to add more unit tests with interesting cases by the end of this week

mosheraboh (approved):

PR #375, opened 2024-10-10
Tokenizer's default behavior on unknown token

SagiPolaczek:
Support a configurable default value for the tokenizer op. Meaning we can change the default behavior of each instance, additionally to override it in the `__call__()` function itself.

### DIFF
--- a/fuse/data/tokenizers/modular_tokenizer/op.py
+++ b/fuse/data/tokenizers/modular_tokenizer/op.py
@@ -30,7 +30,7 @@ class ModularTokenizerWithoutInjectOp(OpBase):
         validate_ends_with_eos: Optional[bool] = True,
         eos: Optional[str] = "<EOS>",
         verbose: Optional[bool] = False,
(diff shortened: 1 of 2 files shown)


fuse/data/tokenizers/modular_tokenizer/op.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         validate_ends_with_eos: Optional[bool] = True,
         eos: Optional[str] = "<EOS>",
         verbose: Optional[bool] = False,
+        on_unknown_default_value: Optional[str] = None,
```
mosheraboh, line 33: Why None? set it directly to "warn".
Lets remove the option to set it to None also in the other function and just support "warn" or "raise"

mosheraboh (commented):
Thanks Sagi. Looks good.
One comment inline.

mosheraboh (approved):
Looks great

PR #376, opened 2024-10-19
Bug fix in `ModularTokenizer.decode()` when the input is `torch.Tensor` type

SagiPolaczek:
Another "solution" would be to validate that the input value is a list of integers.

The main problem in the current implementation, is that the user is not aware of the type mismatch - the token just appears as missing (!)

### DIFF
--- a/fuse/data/tokenizers/modular_tokenizer/modular_tokenizer.py
+++ b/fuse/data/tokenizers/modular_tokenizer/modular_tokenizer.py
@@ -1150,7 +1150,7 @@ class ModularTokenizer(transformers.PreTrainedTokenizerFast):
             str: _description_
         """
         if isinstance(ids, Tensor):
-            # Tokens in 'self.decoder_dict' are integers, and not singletons
+            # Tokens in 'self.decoder_dict' are integers, and not tensors scalars
             ids = ids.tolist()
 
         if skip_special_tokens:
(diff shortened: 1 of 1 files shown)


fuse/data/tokenizers/modular_tokenizer/modular_tokenizer.py
```diff
@@ -1148,6 +1149,9 @@ def decode(self, ids: Iterable, skip_special_tokens: Optional[bool] = False) ->
         Returns:
             str: _description_
         """
+        if isinstance(ids, Tensor):
+            # Tokens in 'self.decoder_dict' are integers, and not singletons
```
mosheraboh, line 1153: singletons?

mosheraboh: Do we need also to move it to CPU before? in cases the tensor is on the GPU.

SagiPolaczek: Good question, so it's being done automatically:
> Returns the tensor as a (nested) list. For scalars, a standard Python number is returned, just like with [item()](https://pytorch.org/docs/stable/generated/torch.Tensor.item.html#torch.Tensor.item). Tensors are automatically moved to the CPU first if necessary.


https://pytorch.org/docs/stable/generated/torch.Tensor.tolist.html

SagiPolaczek: > singletons?

Meaning a tensor with a single item in it. Will make it clearer

mosheraboh (approved):
LGTM

PR #377, opened 2024-10-21
Bump python version for MAMMAL release

SagiPolaczek:


setup.py
```diff
@@ -65,11 +65,11 @@
...
     },
-    python_requires=">=3.7",
+    python_requires=">=3.9",
     classifiers=[
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
+        "Programming Language :: Python :: 3.12",
```
mosheraboh, line 73: I don't think we support 3.12
I think that we should move back to pyproject.toml to support it completely.

mosheraboh (approved):
LGTM

PR #380, opened 2024-11-17
modular tokenizer: add random truncation option

mosheraboh:


fuse/data/tokenizers/modular_tokenizer/modular_tokenizer.py
```diff
@@ -1214,23 +1238,33 @@ def encode(
...
+            curr_truncate_mode = None
+            for part in tokenizer_hints_parts[1:]:
+                if part.startswith("MAX-LEN="):
+                    try:
+                        curr_max_len = int(part[len("MAX-LEN=") :])
+                    except:
+                        raise Exception(
+                            f"Had a problem casting curr_max_len={part} to int! it was found inside modular tokenizer meta TOKENIZER-TYPE={tokenizer_hints}"
+                        )
+                elif part.startswith("TRUNC-MODE="):
```
YoelShoshan, line 1256: can this be added in some documentation somewhere ?

fuse/data/tokenizers/modular_tokenizer/modular_tokenizer.py
```diff
@@ -1214,23 +1238,33 @@ def encode(
...
+                    curr_truncate_mode = part[len("TRUNC-MODE=") :]
+                    assert curr_truncate_mode in [
```
YoelShoshan, line 1258: do you require a certain order between MAX-LEN and TRUNC-MODE?

if not, good.
Asking as a thought for the future, as we add more and more options

YoelShoshan (approved):
LGTM!
see comments

fuse/data/tokenizers/modular_tokenizer/modular_tokenizer.py
mosheraboh: It can be in any order.

fuse/data/tokenizers/modular_tokenizer/modular_tokenizer.py
mosheraboh: Good point - will add to the README file.

mosheraboh: Actually, this all part, about meta tokens and TyoedInput is missing and should be added to README.
I'm opening a task for it.
Maybe we can use the text from the paper to improve the README of the modular tokenizer.

PR #382, opened 2024-11-20
Regression metrics patch

IdoAmosIBM:


fuse/eval/metrics/regression/metrics.py
```diff
@@ -71,6 +96,7 @@ def __init__(
         self,
         pred: str,
         target: str,
+        mask: Optional[str] = None,
```
SagiPolaczek, line 99: `str`? is an error?

fuse/eval/metrics/regression/metrics.py
```diff
@@ -102,6 +152,7 @@ def __init__(
...
```
SagiPolaczek, line 155: str

SagiPolaczek (approved):
LGTM

PR #383, opened 2024-11-21
Allow skipping stats in GroupAnalysis metric

sivanravidos:
GroupAnalysis metrics returns both the metric on each group, and stats (mean, std, median) of these results over the groups. Here we add a parameter to skip the stats, in case the user needs only the groups.


mosheraboh (approved):
LGTM

PR #384, opened 2024-12-05
Sample id fix in GroupAnalysis

sivanravidos:
support tuple ids in GroupAnalysis by converting boolean masking to list comprehension.
Please see #362 for a discussion on a previous solution to his problem in bootstrapping.
This should be reviewed after #383  since it was created over its branch


mosheraboh (approved):
LGTM

PR #390, opened 2024-12-22
Ops imports

YoelShoshan:


SagiPolaczek (approved):
Generally speaking I don't like importing from `__init__.py` files (: 

Anyway, it passes and looks good

PR #393, opened 2025-01-23
fix num_available_cores in lsf

michalozeryflato:


SagiPolaczek (approved):
We miss printing the answer in the case where `verbose is True`.
Not a big deal tho.. I don't like it anyway :P 

LGTM

SagiPolaczek:
Maybe one thing worth mentioning, I think `os.env` returns a string.

SagiPolaczek (approved):
LGTM

PR #394, opened 2025-01-24
Active ranking

YoelShoshan:
helper utilities to rank items based on either:

1. A model that given items a,b predicts whether a should be ranked lower than b
2. A model that given a list of items (e.g. 8 items) predicts their ranking

The user is expected to supply such callable, and the helper functions added in this PR will call it iteratively to predict the overall ranking of the entire provided set.


fuse/eval/metrics/ranking/efficient_active_ranking_pairwise_model.py
```diff
@@ -0,0 +1,164 @@
+import numpy as np
+from typing import Callable, List, Any
+from dataclasses import dataclass
+from collections import defaultdict
```
SagiPolaczek, line 4: Note the `isort` error.

Maybe you need to reinstall the pre-commit hooks? Something like `pre-commit install` when the repo is the current directory. If I right, you'll need to re-run the pre-commits afterwards to make sure it applies the changes :)

SagiPolaczek: Besides that the Jenkins seems to pass (after a rebuilt)

fuse/eval/metrics/ranking/efficient_active_ranking_pairwise_model.py
```diff
@@ -0,0 +1,164 @@
...
+        if method == "merge":
+            return self._adaptive_merge(items)
+        elif method == "quick":
+            self._adaptive_quicksort(items, 0, len(items))
+            return items
+        else:
+            raise ValueError(f"Unknown method: {method}")
+
+
+if __name__ == "__main__":
```
mosheraboh, line 140: Consider converting it to unittest.

fuse/eval/metrics/ranking/efficient_ranking_batch_rank.py
```diff
@@ -0,0 +1,149 @@
...
+                    item_scores[lower_item] -= 1
+                    unique_pairs.add(pair)
+
+    # Sort items based on their aggregate score
+    global_ranking = sorted(all_items, key=lambda x: item_scores[x], reverse=True)
+
+    return global_ranking
+
+
+if __name__ == "__main__":
```
mosheraboh, line 68: consider converting it to unittest

fuse/eval/metrics/ranking/efficient_active_ranking_pairwise_model.py
```diff
@@ -0,0 +1,164 @@
...
+    comparisons: int = 0
+    lower_bound: float = 0.0
+    upper_bound: float = 1.0
+
+
+class EfficientRanking:
+    def __init__(
+        self,
+        items: List[Any],
+        compare_pairwise_fn: Callable[[Any, Any], bool],
```
mosheraboh, line 19: Can you share some details about the expected arguments of this function?

mosheraboh (commented):
Looks good.
General comment, do you use this code to evaluate your model?
I wonder if it indeed should be under eval/metrics?
If it should, I suggest moving it to eval/metrics/lib which is used for lower-level metrics functionality.

YoelShoshan:
> Looks good. General comment, do you use this code to evaluate your model? I wonder if it indeed should be under eval/metrics? If it should, I suggest moving it to eval/metrics/lib which is used for lower-level metrics functionality.

It's a building block to be used. It will not be directly used by any evaluate function.
It's a way for a model builder to come up with ranking (within assay).

The evaluate functions will get as input an already ranked list of items.
It's the responsibility of the model developer to use this building block.

YoelShoshan:
> Looks good. General comment, do you use this code to evaluate your model? I wonder if it indeed should be under eval/metrics? If it should, I suggest moving it to eval/metrics/lib which is used for lower-level metrics functionality.

I see eval/metrics/**libs**  - I'll move the 2 script files to it.

mosheraboh (approved):
Looks great, thanks for also creating a workaround for the CI issue,

PR #399, opened 2025-05-12
add support for modular tokenizer batch encoding

mosheraboh:


fuse/data/tokenizers/modular_tokenizer/modular_tokenizer.py
```diff
@@ -2148,4 +2161,5 @@ def __call__(
                 k: v[0] if isinstance(v, list) and not isinstance(v[0], list) else v[0]
                 for k, v in result.items()
             }
-        return result
+
+        return BatchEncoding(result, tensor_type=return_tensors, encoding=encoding_list)
```
floccinauc, line 2165: Do we need to pass both result and encoding_list here? Won't encoding_list be enough?

floccinauc (approved):
LGTM. A minor remark inline.

PR #402, opened 2025-07-01
adding basic OpCopy class

mosheraboh:


fuse/data/ops/ops_common.py
```diff
@@ -621,6 +621,20 @@ def __call__(
...
+    """copy value from one key to another"""
+
+    def __call__(
+        self, sample_dict: NDict, key_source: str, key_dest: Any
+    ) -> Union[None, dict, List[dict]]:
+        """
+        :param key_source: from where to copy
+        :param key_dest: where to copy to
+        """
+        sample_dict[key_dest] = sample_dict[key_source]
```
YoelShoshan, line 634: Do we want to have some way of actually copying the content?

as it is it looks like it will only create another reference to the same object

PR #409, opened 2025-08-06
edit pyproject.toml-without automatic version

Yishailevine:
There is a conflict between this line and the static version definition:
[project]
version="1.0"


bensha6757 (approved):
LGTM

PR #415, opened 2026-03-04
fix in managing predictions in roc_auc metrics

ellabarkan:
The predictions that were sent to the metrics were generated during mammal finetuning using mammal OS code and were presented as a list and not ndarray. According to the previous logic the case1 was choosen and the metric failed


mosheraboh (approved):
LGTM

